PR #13, opened 2014-12-31
Add StdStream(NoStream) to have standard handles closed.

proger:
The motivation is that lots of processes behave differently when (especially) stdin is
closed vs when it's a pipe.

### DIFF
--- a/System/Process/Internals.hs
+++ b/System/Process/Internals.hs
@@ -223,6 +223,7 @@ data StdStream
                              -- @Handle@ will use the default encoding
                              -- and newline translation mode (just
                              -- like @Handle@s created by @openFile@).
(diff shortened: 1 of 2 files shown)


dcoutts:
Seems reasonable in principle. As @hvr suggests, it can probably wait a bit 'til we need a major version bump anyway.

snoyberg:
Thank you!

proger:
Yay!

snoyberg:
Sorry for the delay. I've been debating pushing forward a more significant
overhaul to make it easier to make changes like this in the future, but
decided such a change was not prudent. There's one more change that I'm
waiting on, and then I'll release this.

PR #21, opened 2015-02-17
Add stopDelegateControlC for MinGW32

takenobu-hs:
Hi,

I would like to thank a lot of effort.
I added stopDelegateControlC function for MinGW32.
Please check:-)

### DIFF
--- a/System/Process/Internals.hs
+++ b/System/Process/Internals.hs
@@ -469,6 +469,9 @@ startDelegateControlC = return ()
 endDelegateControlC :: ExitCode -> IO ()
 endDelegateControlC _ = return ()
 
+stopDelegateControlC :: IO ()
+stopDelegateControlC = return ()
+
 foreign import ccall unsafe "runInteractiveProcess"
   c_runInteractiveProcess
         :: CWString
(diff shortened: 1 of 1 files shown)


snoyberg:
I'm afraid I don't understand the purpose of this patch. Can you describe what the problem is that you are trying to solve, and how this solves it?

hvr:
@snoyberg, @thoughtpolice will chime in soon; this is about https://mail.haskell.org/pipermail/ghc-devs/2015-February/008347.html

thoughtpolice:
3b5804ebe54c672146994cfc525d45280650043e introduces the use of `stopDelegateControlC` inside `Process.hsc`. This lives in `Internals.hs`, which has an export list that always exports this function. But the implementation is `#ifdef`'d between Unix and Windows. This causes a build failure on Windows where `stopDelegateControlC` was not defined, which this patch fixes (it should be a no-op on Windows). I believe Takenobu's patch is the easiest way to fix this.

snoyberg:
Thanks for the clarification, that makes sense.

@takenobu-hs Would you be able to update this PR to include a short explanation in the code as to why we have that no-op function? It could even just be a link pointing to this discussion.

takenobu-hs:
Sorry less information and thank you for the detailed instructions.

I updated the commit message and re-sent a PR for issue #21
If I mistake something, please teach again.

Thank you.

snoyberg:
I was actually hoping for the comments to be added in the code, not the commit message, so that in the future, when someone is reading through the codebase, they can understand why there's a no-op function present. Can you amend in that way?

takenobu-hs:
Oh, I understood.
Thank you for repeatedly teach me.

I added a comment in the code.

How about this?

snoyberg:
Thanks!

PR #24, opened 2015-02-25
Add readCreateProcess function

niteria:
This function is more flexible then readProcess, for example it allows
you to change the working directory before running the command.

This loses a bit in terms of error reporting. I'm also not sure if extending the API is the right thing to do. Another way to achieve what I want would be to expose withCreateProcess_, withForkWait, ignoreSigPipe, processFailedException functions in System.Process.Internals.

### DIFF
--- a/System/Process.hsc
+++ b/System/Process.hsc
@@ -422,6 +422,8 @@ readProcess cmd args = readCreateProcess $ proc cmd args
 -- >  > readCreateProcess (shell "pwd" { cwd = "/etc/" }) ""
 -- >  "/etc\n"
 --
+-- Note that @Handle@s provided for @std_in@ or @std_out@ via the CreateProcess
+-- record will be ignored.
 -- /Since: 1.2.3.0/
 
 readCreateProcess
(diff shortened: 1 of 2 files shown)


hvr:
(minor nitpick: changelog.md entry missing as well as a `@since`-annotation...)

snoyberg:
@niteria Can you update the PR to reflect @hvr's comments?

niteria:
Not sure if 1.2.3.0 is the right version.

System/Process.hsc
```diff
@@ -413,13 +414,27 @@ readProcess
...
+-- | @readCreateProcess@ works exactly like 'readProcess' except that it
+-- lets you pass 'CreateProcess' giving better flexibility.
+--
+-- >  > readCreateProcess (shell "pwd" { cwd = "/etc/" }) ""
+-- >  "/etc\n"
+--
+-- /Since: 1.2.3.0/
+
+readCreateProcess
+    :: CreateProcess
```
snoyberg: I think it's worth commenting that any values for `std_in`, `std_out` and `std_err` will be overwritten. Additionally: is it worth including `readCreateProcessWithExitCode` as well?

niteria:
I haven't found a need for `readCreateProcessWithExitCode` so far, but I wouldn't mind adding it.
I'll update the comment.

snoyberg:
LGTM, thanks!

PR #26, opened 2015-03-08
Update 2 GHC tests

thomie:
The first commit is needed for `process master` to pass GHC validation, after the recent addition of `readCreateProcess`.

The second commit makes the `process` testsuite compatible with some changes I have planned for the GHC test framework.


snoyberg:
Thanks!

PR #33, opened 2015-05-27
add detach_console flag

### LINKED ISSUE #32: expose windows DETACHED_PROCESS flag
On Windows the DETACHED_PROCESS flag lets the process detached from the terminal. I need this to run ssh with SSH_ASKPASS, which only works when it has no controlling console. So, could this be added as a new field to CreateProcess, perhaps called detach_from_console?

I needed to set both DETACHED_PROCESS and CREATE_NO_WINDOW to get the desired result. (Otherwise, windows pops up a new console window.)

It might make sense to have the unix implementation of detach_from_console call setsid.

If this plan seems to make sense, I'll cook up a patch, since I need this functionality.

joeyh:
Based on Daniel Brooks's changes, but without adding a new flag
for CREATE_NO_WINDOW, which turned out to not be necessary.
And, on unix, setsid is called.

Tested on Linux and Windows.

Fixes #32.


snoyberg:
I'm still of the opinion that these need to be two separate flags, the reason being: with two flags, getting any combination of behavior is trivial, but with a single flag, users are forced into difficult to use conditional compilation.

PR #39, opened 2015-08-25
1.3 changelog: unreleased -> August 2015

rvion:
minor tweak, may not worth merging

/cc @snoyberg


snoyberg:
Doh, good catch, thanks! Very much worth merging actually

PR #41, opened 2015-09-19
Add blank comment before "@since 1.2.3.0"

kraai:
"@since 1.2.3.0" is included verbatim in the last paragraph of readCreateProcess's documentation.  I think this occurs because there isn't a blank comment separating it from the last paragraph.


snoyberg:
Thanks!

PR #43, opened 2015-09-29
Join more sentence fragments

kraai:
Sorry for not fixing this in my previous commit.  I didn't realize the same problem was repeated.  :(


snoyberg:
Thanks again!

Issue #44, opened 2015-10-08
Feature request: Ability to change user and group


jprider63:
I'd like to change the user and group of the forked process (I'm using Linux, but I think this would still make sense for Windows). I looked into implementing this myself, but it seems like there's extensive FFI use (ie `runInteractiveProcess`), and it could require some potentially breaking API changes. 

Do you think this is feasible?

snoyberg:
When I've needed this I've always used `sudo`. I suppose adding some logic to the C part of the code to optionally call setuid and setgid should work. I can't think of a way to make this meaningful on Windows, however.

jprider63:
I'm currently using a script to change the group and user, which then calls the target executable. I avoided using sudo because I need to explicitly control argv inputs and was concerned about the shell having to parse the arguments. 

For the API, maybe we could add two fields to `CreateProcess` called `child_user :: Maybe String` and `child_group :: Maybe String`. I could take a shot at implementing this if it seems reasonable to you.

snoyberg:
Seems fine. My only question would be whether it should use a String or
expect Int (user ID). I'm fine either way.

Please note in the haddocks which OSs it's expected to work on.

jprider63:
I just made a pull request with this implemented. I ended up going with `CGid` and `CUid` instead of `String`. I've only tested on unix with ghc, but I think other targets should be unaffected.

snoyberg:
Implemented by PR #45, thank you!

PR #45, opened 2015-10-14
Added child_user and child_group to CreateProcess for unix.

jprider63:
Implements #44.


System/Process/Internals.hs
```diff
@@ -190,9 +190,19 @@ data CreateProcess = CreateProcess{
...
                                            --   @since 1.3.0.0
-  new_session :: Bool                      -- ^ Use posix setsid to start the new process in a new session; does nothing on other platforms.
+  new_session :: Bool,                     -- ^ Use posix setsid to start the new process in a new session; does nothing on other platforms.
                                            --
                                            --   @since 1.3.0.0
+  child_group :: Maybe GroupID,            -- ^ Use posix setgid to set child process's group id; does nothing on other platforms.
+                                           --
+                                           --   Default: @Nothing@
+                                           --
+                                           --   @since X.X.X.X
```
snoyberg: Can be set at 1.4.0.0

System/Process/Internals.hs
```diff
@@ -414,6 +429,8 @@ foreign import ccall unsafe "runInteractiveProcess"
         -> Ptr FD
         -> Ptr FD
         -> Ptr FD
+        -> Ptr CGid
```
snoyberg: Given that a CGid or CUid is just an `int`, I think this can be done without the pointer indirection, by using `-1` to represent "not provided."

jprider63: I've made this change in the latest commit, but I'm worried that there could be conflicts since `uid_t` and `gid_t ` are unsigned. It might be better to go with the original.

snoyberg: The research I did before making my comment implied that:

1. Signedness is implementation specific, and will vary by platform
2. Even on platforms with an unsigned type, the `-1` value is considered reserved

I could be mistaken on this, however. I'm OK with either approach, it just struck me as strange to use a pointer for integral data.

jprider63: I agree that signedness will vary by platform. Do you have a reference that says `-1` is reserved? My concern was that on an unsigned system, `-1` could be interpreted as max_int (which although unlikely could be a user's/group's id). 

I suppose another option would be to add another mask to the flag indicating whether the set the user/group or not. Maybe this is the safest approach.

snoyberg: What I've found:

http://pubs.opengroup.org/onlinepubs/009695399/functions/chown.html
http://stackoverflow.com/questions/21370094/uid-t-type-is-signed-or-not

I wouldn't bother with a mask on the flag, I'd either stick with your original pointer approach or use `-1`.

jprider63: It looks like using `-1` is probably ok, but I'm still skeptical since [setgid's](http://pubs.opengroup.org/onlinepubs/009695399/functions/setgid.html) and [setuid's](http://pubs.opengroup.org/onlinepubs/009695399/functions/setuid.html) documentation don't mention it. I'll revert to the pointer approach since I think it's the safer option.

snoyberg:
Thanks!

PR #47, opened 2015-11-03
Use less CPP

snoyberg:
Major changes:

* Most Windows-specific and POSIX-specific code live in their own modules
* No more GLASGOW_HASKELL conditionals, since it didn't work anyway

This now compiles without warnings on Linux and Windows, addressing Simon's original concern.

@hvr would you mind reviewing?


System/Process.hsc
```diff
@@ -1,12 +1,10 @@
 {-# LANGUAGE CPP, ForeignFunctionInterface #-}
-#ifdef __GLASGOW_HASKELL__
```
Mistuke: Shouldn't this still be here? Unless `InterruptibleFFI` is a portable extension now? or process only supports GHC?

process.cabal
```diff
@@ -44,9 +44,16 @@ library
...
-    if impl(ghc)
-        exposed-modules:
-            System.Process.Internals
+        System.Process.Internals
+    other-modules: System.Process.Common
+    if os(windows)
+        other-modules: System.Process.Windows
+        build-depends: Win32 >=2.2 && < 2.4
+        extra-libraries: kernel32
+        cpp-options: -DWINDOWS
```
hvr: Can't we just use the official macros defined by GHC?

Quoting the documentation:

```
``os_HOST_OS=1``
    This define allows conditional compilation based on the Operating
    System, where⟨os⟩ is the name of the current Operating System (eg.
    ``linux``, ``mingw32`` for Windows, ``solaris``, etc.).

``arch_HOST_ARCH=1``
    This define allows conditional compilation based on the host
    architecture, where⟨arch⟩ is the name of the current architecture
    (eg. ``i386``, ``x86_64``, ``powerpc``, ``sparc``, etc.).
```

snoyberg: Usage was inconsistent in the codebase previously around a few different ways of detecting Windows, and I've seen concerns raised in the past about the reliability of those macros. I feel more comfortable using the same logic for the Haskell code conditionals as the Cabal conditionals guiding the package dependencies and module selection.

System/Process/Common.hs
```diff
@@ -0,0 +1,215 @@
...
+    , fd_stdin
+    , fd_stdout
+    , fd_stderr
+    , mbFd
+    , mbPipe
+    , pfdToHandle
+
+-- Avoid a warning on Windows
+#ifdef WINDOWS
+    , CGid (..)
```
Mistuke: The only comment I really have is still about this. I have read the emails, and I think you're primary motive for doing it like this is so the APIs for Windows and Linux don't diverge. But don't they already?

The `CGid` for Linux seems to be a  type alias (I don't know where it comes from so can't check). I think by exporting the constructor on Windows it's actually giving the opposite impression. To be, without knowing the back story here it looks like it is *actually* being used on windows.

If you do like it this way better, consider Simon's suggestion of adding a `{-# WARNING #-}` so that if someone does use it, they get a warning.

snoyberg: I think we're on the same page: I'd rather _not_ export the `CGid` data constructor. That's what you're saying too, right? In fact, that's what this PR does for the public-facing API. The .Common module is not exported from the package, and the .Internals module does not export the `CGid` data constructor.

Mistuke: Ah I see, you've moved the module to hide it from the Users of process but not the developers. Alright then.

PR #52, opened 2016-01-24
Trac 11100 expose part of create pipe

Mistuke:
Would like to expose a part of the `createPipe` api which just returns an `Posix.FD` instead of a GHC `Handle`. 

This is for the implementation of the Windows bit of `RemoteGHCI` https://ghc.haskell.org/trac/ghc/ticket/11100


snoyberg:
This PR works on Windows, but not on POSIX systems.

Mistuke:
Ah, I'll fix that. Thanks!

Mistuke:
@snoyberg I think the appveyor one failed because of a network failure?

System/Process/Internals.hs
```diff
@@ -163,6 +165,17 @@ createPipe :: IO (Handle, Handle)
 createPipe = createPipeInternal
 {-# INLINE createPipe #-}
 
+-- ---------------------------------------------------------------------------
+-- createPipeFD
+
+-- | Create a pipe for interprocess communication and return a
+-- @(readEnd, writeEnd)@ `FD` pair.
+--
+-- @since 1.2.1.0
```
snoyberg: This should actually be 1.4.2.0. Similarly, can you update the ChangeLog and the .cabal file?

System/Process/Internals.hs
```diff
@@ -38,6 +38,7 @@ module System.Process.Internals (
     withFilePathException, withCEnvironment,
     translate,
     createPipe,
+    createPipeFD,
```
bgamari: It seems common to abbreviate "file descriptor" as `Fd`. Given that this convention is used below perhaps it would make sense to preserve consistency here?

bgamari:
Thanks! @snoyberg, would you like to merge this or shall I?

snoyberg:
I've got it, thanks Ben. I'll make a release to Hackage now.

bgamari:
Thanks Michael!

hvr:
@snoyberg btw, please don't forget to `git tag -a -s v1.4.2.0 ...` + `git push v1.4.2.0` =)

snoyberg:
Done Herbert, thanks for the reminder.

Issue #59, opened 2016-04-06
Binary I/O versions of readProcess, readProcessWithExitCode, etc.


robjhen:
It would be useful to have versions of readProcess etc. that support binary IO on stdin/stdout.
Proposed solution: a new module (intended to be imported qualified alongside System.Process) with the following signature:

``` haskell
module System.Process.Binary where
import Data.ByteString.Lazy (ByteString)
readProcess :: FilePath -> [String] -> ByteString -> IO ByteString
readCreateProcess :: CreateProcess -> ByteString -> IO ByteString
readProcessWithExitCode :: FilePath -> [String] -> ByteString -> IO (ExitCode,ByteString,ByteString)
readCreateProcessWithExitCode :: CreateProcess -> ByteString -> IO (ExitCode,ByteString,ByteString)
```

Additionally, I propose adding a new constructor CreateBinaryPipe to the StdStream data type (CreateBinaryPipe would be to CreatePipe as openBinaryFile is to openFile).

If there are no objections, I'd be happy to implement this myself and submit a patch. Note that it would require adding bytestring as a package dependency.

snoyberg:
This was discussed previously in #36.

robjhen:
Thanks for pointing out #36, which I hadn't spotted. I see that you got quite close to adding this feature but then a decision was made to drop it due to the presence of process-extras.

However, I'd like to try giving this another shot if it's okay. In my opinion, process-extras is not a satisfactory solution for the following reasons:

 1) The process-extras implementation of readProcess (et al.) duplicates code that is already in 'process'. This makes maintenance difficult because any bug fix or other change in the 'process' version of readProcess will not automatically make its way into 'process-extras' and vice-versa.

 2) process-extras and a number of its dependencies are outside of the Haskell Platform. I think that a binary version of readProcess is important enough that it really ought to go in the Haskell Platform (I argue a case for this below).

### Why this matters: a motivating example

Consider a typical everyday scripting task: computing the SHA256 hash of an SSH public key. With a binary version of readProcess this can be done elegantly as follows:

```haskell
example :: IO ()
example =
  -- Extract base64 public key from file
  fmap ((!! 2) . words) (readFile "key.pub")
  >>=
  -- Convert from base64 to binary
  B.readProcess "base64" ["-d"] . toUtf8
  >>=
  -- Compute sha256sum and convert from hex to binary
  fmap (fromHex . head . words . fromUtf8) .
       B.readProcess "sha256sum" ["-b"]
  >>=
[...]

PR #61, opened 2016-04-27
#59 (binary versions of readProcess, etc.)

robjhen:
I have structured this pull request as 3 separate commits for ease of reviewing - see commit messages for details.

Notes:
- Tested on Debian GNU/Linux with both ghc 7.4.1 and ghc 7.10.2.
- Was unable to test on Windows as I do not have access to a Windows machine.
- I have not given an upper bound for the 'bytestring' package dependency in process.cabal: you may wish to add one. However I chose the lower bound (0.9.2) with some care: it is the lowest value compatible with the other package constraints, and it is also the version of bytestring that comes with ghc 7.4.1.


robjhen:
The automated build failures appear to be because Hackage was down for maintenance at the time? The Travis build reported "cabal update failed 3 times" then "unknown package: HUnit". Is there a straightforward way to retry these builds?

ggreif:
you can restart them in travis

ggreif:
@robjhen I restarted the builds. `cabal` did well, `stack` not so.

robjhen:
Okay, all done. The build failures with stack were due to a dependency on HUnit for a test suite added by me. I fixed this problem by changing the resolver in stack.yaml from ghc-7.10.2 to lts-5.14, and making some related changes in the .travis.yml script.

snoyberg:
I'm concerned about such a large change, on a feature that was controversial to add to process in the first place. I'm also not sure what advantages are gained from copying so much code around instead of simply using `System.Process` exported identifiers in the binary module.

Some of the work here is straightforward and can be merged in, like changing the .hsc file extension and cleaning up language extensions. Can we start with a PR to just do that? There are then other breaking changes that may be worth doing, but need a little bit more exploration (CreateBinaryPipe caught my eye). After those are taken care of, we can circle back to considering the main functionality additions. But my preference would be for more focused code additions rather than large restructurings.

robjhen:
Okay, agreed - it makes sense to approach this in smaller chunks. I'll make a new pull request for changing the .hsc extension.

PR #62, opened 2016-05-12
Removed .hsc extension no longer in use

robjhen:
I've removed the .hsc extension from Process.hs but left the one on Process/Windows.hsc which is still required. As for language extensions, I believe they are all still in use.


robjhen:
The build failure was due to some kind of glitch in one of the jobs, nothing to do with the commit:
`curl: (35) Unknown SSL protocol error in connection to github-cloud.s3.amazonaws.com:443 
`

snoyberg:
Thanks!

PR #63, opened 2016-05-23
Issue #59: added CreateBinaryPipe constructor to StdStream.

robjhen:
CreateBinaryPipe is to CreatePipe as openBinaryFile is to
openFile. See Process/Common.hs for the implementation.

I also added a test suite (test-binary.hs). I had to change the
stack resolver from ghc to lts in order to allow an HUnit
dependency.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,20 +36,20 @@ matrix:
     compiler: ": #GHC 7.10.3"
     addons: {apt: {packages: [cabal-install-1.22,ghc-7.10.3], sources: [hvr-ghc]}}
 
-  - env: BUILD=stack GHCVER=7.8.4
-    compiler: ": #stack 7.8.4"
+  - env: BUILD=stack LTS_HASKELL_VER=2.22
+    compiler: ": #stack lts-2.22"
     addons: {apt: {packages: [ghc-7.8.4], sources: [hvr-ghc]}}
 
-  - env: BUILD=stack GHCVER=7.10.3
-    compiler: ": #stack 7.10.3"
+  - env: BUILD=stack LTS_HASKELL_VER=5.14
+    compiler: ": #stack lts-5.14"
     addons: {apt: {packages: [ghc-7.10.3], sources: [hvr-ghc]}}
 
-  - env: BUILD=stack GHCVER=7.8.4
-    compiler: ": #stack 7.8.4 osx"
+  - env: BUILD=stack LTS_HASKELL_VER=2.22
+    compiler: ": #stack lts-2.22 osx"
     os: osx
 
-  - env: BUILD=stack GHCVER=7.10.3
-    compiler: ": #stack 7.10.3 osx"
+  - env: BUILD=stack LTS_HASKELL_VER=5.14
+    compiler: ": #stack lts-5.14 osx"
     os: osx
 
   - env: BUILD=cabal GHCVER=head  CABALVER=head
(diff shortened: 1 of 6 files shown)


robjhen:
The AppVeyor build failed trying to download prerequisite libraries from 'downloads.sourceforge.net'. This looks like an issue with a server being down and not a problem with the commit.

snoyberg:
There isn't a justification given for why this new constructor is useful, and the test case is very contrived. Can you explain why this change is necessary?

robjhen:
CreateBinaryPipe lets you open a pipe to subprocess directly in binary mode, rather than having to open it in text mode first and then change to binary mode using hSetBinaryMode. This has two specific advantages:

1. **Consistency with System.IO**. In System.IO, operations that return new handles come in text/binary mode pairs: openFile/openBinaryFile, withFile/withBinaryFile, openTempFile/openBinaryTempFile, etc. However, the 'CreatePipe' feature of createProcess, which returns a handle to a pipe, currently only supports text mode directly and is lacking its binary mode counterpart.

2. **Efficiency**. Using CreateBinaryPipe is more efficient than using CreatePipe followed by hSetBinaryMode, as the latter involves a needless allocation and deallocation of TextEncoder/TextDecoder objects as well as a buffer flush.

snoyberg:
> CreateBinaryPipe lets you open a pipe to subprocess directly in binary mode, rather than having to open it in text mode first and then change to binary mode using hSetBinaryMode.

But that's ignoring the main thrust of my question: why do we need a pipe in binary mode? For the use cases we're discussing here, what errors would a text mode pipe introduce?

robjhen:
As an example use case, suppose we want to compress a piece of data by invoking the 'gzip' program as a subprocess. The compressed data will be non-textual, so we'll need to read it from gzip's stdout pipe as a bytestring, e.g. using Data.ByteString.Lazy.hGetContents.

From the [Haddock documentation of Data.ByteString.Lazy.hGetContents](http://hackage.haskell.org/package/bytestring-0.10.8.1/docs/Data-ByteString-Lazy.html#v:hGetContents):

"the Handle should be placed in binary mode ... for hGetContents to work correctly."

So, the bytestring documentation is telling us that if we try to read a bytestring from a pipe in text mode, potentially undefined behaviour may result. Hence the need to open pipes in binary mode.

snoyberg:
That's the root of my confusion: I think that comment in the bytestring docs is incorrect. For example, with this program, I do not see any carriage return manipulation happening. In other words: AFAICT, all bytestring-based I/O on Handles ignores the binary mode:

```haskell
{-# LANGUAGE OverloadedStrings #-}
import qualified System.IO as IO
import qualified Data.ByteString as S

main :: IO ()
main = do
     let input = "line 1\nline2\r\nline3\rline4"
         fp = "tmp"
     S.writeFile fp input
     IO.withFile fp IO.ReadMode $ \h -> do
       IO.hSetNewlineMode h IO.universalNewlineMode
       S.hGetContents h >>= print
```

Is there some case you're aware of that doesn't work like this?

robjhen:
Right, now I see what you mean. Before we make a decision, I'll drop the maintainer of the bytestring package a quick e-mail just to get confirmation on this point.

snoyberg:
Note that in https://github.com/haskell/process/pull/67 I added a test for this, and it seems to work on all OSes.

snoyberg:
@robjhen Any follow-up from the bytestring maintainer?

robjhen:
I didn't hear back by e-mail; perhaps it's better to contact via github. I've just opened an [issue in the bytestring package](https://github.com/haskell/bytestring/issues/97) to get confirmation about whether that line in the bytestring Haddock docs is out of date.

ndmitchell:
I would have found this facility useful. In https://github.com/ndmitchell/hoogle/issues/194 I needed to switch from readProcess with text handles, to readProcess with binary handles. Doing that on top of the process library was a nightmare - you have to rewrite all the nasty stream-draining/exiting code yourself. I ended up using the process-bytestring library instead. If this patch was in I believe I could have switched to `readCreateProcessWithExitCode` passing `CreateBinaryPipe`, provided `readCreateProcessWithExitCode` didn't clobber the pipe values (which I don't think it should).

jml:
> Is there some case you're aware of that doesn't work like this?

The case that @ndmitchell points out: extended characters (such as `é`) in stdout streams on systems where you cannot guarantee correct `LANG` settings.

snoyberg:
@jml I don't believe you've read my previous comments, your comments don't seem relevant to the topic of the bytestring package.

snoyberg:
Closing, see: https://github.com/haskell/process/issues/59#issuecomment-295175436

PR #65, opened 2016-05-26
Close pipes on failure

### LINKED ISSUE #64: createProcess leaks pipes on failure
[Here we create a pipe](https://github.com/haskell/process/blob/296535ef2d1b0d6051a1db8adf402549acd87557/cbits/runProcess.c#L78) for `stdin` (and similar for `stdout` and `stderr`.) In most cases it is correctly closed on failure, e.g. [when `fork` fails](https://github.com/haskell/process/blob/296535ef2d1b0d6051a1db8adf402549acd87557/cbits/runProcess.c#L122). But if `exec` fails (e.g. the executable doesn't exist), we [wait for the child](https://github.com/haskell/process/blob/296535ef2d1b0d6051a1db8adf402549acd87557/cbits/runProcess.c#L329), but don't close the pipes as I'm sure we should.

To reproduce: call `createProcess` with non-existing executable and `stdin` configured as `CreatePipe`. Do it multiple times and monitor number of open FDs in `/proc/<pid>/fd`, it grows each on each `createProcess`.

Yuras:
Make sure stdin, stdout and stderr pipes are closed when exec call
fails, e.g. when the executable doesn't exist.

It is an attempt to fix #64. The code around is pretty complicated, I'm not 100% sure I'm doing everything right, so please review carefully. Anyway, it fixes the immediate issue I had in my application.

### DIFF
--- a/cbits/runProcess.c
+++ b/cbits/runProcess.c
@@ -84,6 +84,10 @@ runInteractiveProcess (char *const args[],
     if (fdStdOut == -1) {
         r = pipe(fdStdOutput);
         if (r == -1) {
+            if (fdStdIn == -1) {
+                close(fdStdInput[0]);
+                close(fdStdInput[1]);
+            }
             *failed_doing = "runInteractiveProcess: pipe";
             return -1;
         }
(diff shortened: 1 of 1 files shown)


snoyberg:
LGTM. @dcoutts or @simonmar, would one of you mind giving this code a once over for sanity?

snoyberg:
@simonmar This patch looks good to me, I'm going to merge it next week unless I hear otherwise.

snoyberg:
Thanks!

PR #69, opened 2016-08-11
Expose withCreateProcess to users.

JonCoens:
The function was commented out and not used anywhere.

It's currently rather cumbersome to run a process in the background, perform some action, and close the process once your action is completed. I believe withCreateProcess can handle that use-case and just wasn't exposed.


snoyberg:
LGTM!

PR #73, opened 2016-10-23
Fix CPP usage

erikd:
The code `#if WINDOWS` works but is not really correct. GHC HEAD now
has a `-Wcpp-undef` warning that we would like to turn on and hence
need this fixed.


snoyberg:
Thanks! Do you need a release of this, or is just merging it sufficient?

erikd:
Merging is sufficient. Thanks!

PR #74, opened 2016-11-11
Add a remark on how relative cmdspec is resolved w.r.t. to cwd.

ezyang:
Signed-off-by: Edward Z. Yang <ezyang@cs.stanford.edu>

### DIFF
--- a/System/Process/Common.hs
+++ b/System/Process/Common.hs
@@ -64,7 +64,7 @@ type PHANDLE = CPid
 #endif
 
 data CreateProcess = CreateProcess{
-  cmdspec      :: CmdSpec,                 -- ^ Executable & arguments, or shell command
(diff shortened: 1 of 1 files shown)


snoyberg:
I'll need to check this next week, ideally with a test case. IIRC, this behavior differs between OSs.

ezyang:
If it is true that the behavior varies across platforms, then there needs to be an admonition that it is invalid to pass relative paths to a cmdspec (because you cannot depend on path resolution behavior to be consistent.)

snoyberg:
Testing indicates that your comments are accurate on all three major OSes, so either the bug was fixed, or I'm misremembering. Either way, improved docs and better test cases are clearly a win here, so thank you for poking at this. I'll merge your changes into my WIP branch and then merge to master after successful tests on all OSes from build servers.

Issue #77, opened 2016-11-30
WaitForSingleObject on Windows isn't sufficient to wait for Process termination.


Mistuke:
Usually when `CreateProcess` is used then `WaitForSingleObject` on the handle as is done here is enough to wait for the process (and it's child processes that it may spawn) to terminate.

The problems arise when inside the child process a new process is spawned using `_execv` or any of the other `_exec` functions. Inside msvcrt `execv` is implemented as `spawnve (_P_OVERLAY, path, argv, cur_environ ())`

`_P_OVERLAY`  is crucial, as it makes the function behave *sort of* like execv on linux. the child process replaces the original process.

With one major difference because of the difference in process models on Windows: the original process signals the caller that it's done.

This is because a call to `CreateProcess` is made followed directly by an `_exit(0)` call. So control is returned to you and you get a success exit code but the child is still running. 

This is what I believe is causing `hsc2hs` to fail on removing an output file:

`Failed to remove file ./T3837_hsc_make; error= DeleteFile "./T3837_hsc_make": permission denied (Access is denied.)`

And requiring this hack which sorta, sometimes works https://github.com/ghc/hsc2hs/blob/f5ae016e5a69ebf42d612805e51afd9227df9389/Common.hs#L53

The solution for this is to use an I/O Completion port and a Job objection to monitor the process tree and wait for all processes in the created job to finish. I can whip up a patch for this but I wanted to see how you guys wanted this to be done.

Do you want a new API, or modify the existing one to use completion ports to wait instead of `WaitForSingleObject`. I think the overhead would be very slight, but it would be more correct in the presence of applications that use the POSIX spawn functions.

snoyberg:
Personally, I'd defer to your best judgement. My guess is that it's worth taking a small performance hit to improve correctness, and if there's a real performance argument to be made, provide the current functionality under a new name which explicitly calls out the issue.

It sounds like the problem you're reporting may be one I accidentally discovered in the past: https://github.com/fpco/replace-process/issues/2. If that's the same bug, this may be a good test case to add.

Mistuke:
Ahh, yes it's the same bug. The patch can fix it as well as you get to see all the exit codes of all child processes. I'll use it as a test case. Thanks!

snoyberg:
Thank you!

PR #80, opened 2016-12-11
Add new API that can correctly wait for termination of processes forked with exec on Windows.

Mistuke:
See https://github.com/haskell/process/issues/77

In the end I had to make a new API, since the handle returned by the Job should never be waited on. 
Existing code would probably call the old wait function on it and then block indefinitely (or until a timeout if any was specified).

Instead I have created a new API and replaced the internals functions that do a fork/exec/wait with the new api.

I have also fixed the issue with the exit code from processes forked with `exec` not correctly being returned.


System/Process.hs
```diff
@@ -870,6 +869,22 @@ It will therefore behave more portably between operating systems than 'system'.
...
 -}
 rawSystem :: String -> [String] -> IO ExitCode
-rawSystem cmd args = do
-  (_,_,_,p) <- createProcess_ "rawSystem" (proc cmd args) { delegate_ctlc = True }
+rawSystem cmd args = executeAndWait "rawSystem" (proc cmd args) { delegate_ctlc = True }
+
+-- ---------------------------------------------------------------------------
+-- executeAndWait
+
+-- | Create a new process and wait for it's termination.
```
simonmar: "its"

simonmar: Please explain how this is different from `createProcess` followed by `waitForProcess`.  It's quite worrying to have two APIs for the same thing that have subtly different behaviour, but I didn't follow all the details of the problem.  Is there really no way to make `waitForProcess` do the right thing?

Mistuke: I really should stop making this typo...

Mistuke: Right, so the basic problem is this:
Say you have two programs `A` and `B`.  Now `A` calls `B`. There are two ways to do this.

1) You can use the normal `CreateProcess` API, which is what normal Windows code do. Using this approach, the current `waitForProcess` works absolutely fine.
2) You can call the emulated POSIX function `_exec`, which of course is supposed to allow the child process to replace the parent. 

With approach `2)` `waitForProcess` falls apart because the Win32's process model does not allow this the same way as linux. `_exec` is emulated by first making a call to `CreateProcess` to spawn `B` and then immediately exiting from `A`. So you have two different processes.

`waitForProcess` is waiting on the termination of `A`. Because `A` is immediately killed, `waitForProcess` will return even though `B` is still running. This is why for instance the testsuite on Windows had lots of file locked errors.

This approach creates a new `Job` and assigned `A` to the job, but also all future processes spawned by `A`. This allows us to listen in on events, such as, when all processes in the job are finished, but also allows us to propagate exit codes from `_exec` calls.

The only reason we need this at all is because we don't interact with just actual native code on Windows, and instead have a lot of ported POSIX code.

I could add a new constructor to `ProcessHandle` and make the decision on what to do internally in the functions. This would probably generate warnings in user code that deconstruct `ProcessHandle` though.

snoyberg: Just to confirm my understanding: with the current API on Linux, if we have:

1. Process A calls `rawSystem "B" []`
2. Process B calls `createProcess $ proc "C" []` and then exits

Then the `rawSystem` call in `A` will return when `B` exits, even if `C` is still running. With this new API on Windows, `A` would block until both `B` and `C` exit. Is this correct?

simonmar: @Mistuke that's a really good explanation, thanks. (it would be great to include this as a comment in the code)

> I could add a new constructor to ProcessHandle

But how would you know when to do that?  Would this entail creating the Job for every process we spawn?

Mistuke: @snoyberg By default, yes. However `CreateProcess` has a flag `CREATE_BREAKAWAY_FROM_JOB` that allows a child process to break away from any parent job. So you can get the same behaviour as on linux by `B` using this flag when it spawns `C`. Though I agree, it is subtle.. (There's also a flag to prevent any children from breaking away, so it goes both ways :))

@simonmar I'll update the comment. 
So there are two alternate approaches that could be taken here:
1) always use a job. In which case, if a user hasn't been deconstructing `ProcessHandle` then it should work transparently.
2) Combine the new function with the old one instead of having two calls now, and make the behaviour implicit on flag. But this would not be backwards compatible, since of course there's a new parameter.

I went with separate functions to avoid always using a job and breaking backwards compatibility. 

If we go for 1) then, according to some very statistically unsound testing, the added overhead is about `0.0021 ms`.

Mistuke:
Also, I found a bug with this version when doing a GHC bootstrapping build. I'll fix that soon.

System/Process.hs
snoyberg: Given the change in behavior, and the fact that we cannot replicate the same behavior on POSIX systems as we do on Windows, I can't imagine a way in which we can add this new functionality to the default `createProcess` settings without wide-spread breakage. We _could_ add a new flag to `CreateProcess` to indicate if this logic should be used. However, this behavior is different enough, and applied to a narrow enough use case (`exec` usage on Windows) that I'd probably stick with a separate function and make very clear the narrow window\* for which this is intended and the caveats that apply.

\* No pun intended

Mistuke: Should note, there's also one useful thing jobs give you: it allows you to get and set detailed resource limits on processes. e.g. not use more than `x` memory, don't run longer than `y` ms, etc.

Which is why I wanted to expose the job handle back to the user as well. Since while not provided here, the changes you would need to set these should be rather trivial now, this is what they are in general used for along with just monitoring processes.

Mistuke: Hmm so both of you seem to favor different options. How about a middle ground. I keep two separate functions for spawning, but only have one for waiting?

snoyberg: Yes, @simonmar and I _do_ seem to be at odds. Simon: AFAICT, switching `createProcess` to have the new job-based behavior is a significantly breaking semantic change on Windows only, which is a huge recipe for disaster. Do you disagree with this assessment?

simonmar: Yes, I see.  What I'm concerned about is adding a new API that duplicates existing functionality but with subtly different behaviour on one platform.  That's also a recipe for disaster.

What is using `exec` on Windows?

Mistuke:
So,  most blatant user at the moment are any msys utilities. Which we call
from ghc. I've patched for instance timeout, but the amount of usages
throughout I figured I should stop duplicating the code. Gcc also uses this
when calling the compiler from the driver. I believe the MINGW guys patch
these  out to CreateProcess but not sure. Calling gcc for compilation seems
to work or we'd have problems already. Calling it for preprocessing seems
to subtly not work properly like in hsc2hs, again I can patch it in the
driver,  but I'm duplicating code everywhere.

This also has a difference in the exit code. You'll get the exit code for
the last process to exit. So if using exec you get the right exit code now.
And when using CreateProcess since the last application would be the first
caller.

Is it really that confusing to have a 'createProcessWithJob' call?

snoyberg:
What about the alternative of having a setting in `CreateProcess` that switches to the job-based behavior, with the clear documentation that:

* It only affects Windows, and has no impact on other OSes
* It does not guarantee identical behavior to other OSes, but instead will have slightly different exit code behavior as we've discussed here

I think that will address the API proliferation issue (which I agree is a problem), and should hopefully be about as easy to work with from a user perspective.

Mistuke:
That works for me as well. So then we're back to one function. And I assume
we just leave the processes that do an execute and wait as they are now,
such as `rawSystem`.

snoyberg:
That's my preference. There's been a move over the past few releases to make `CreateProcess` an easier-to-use part of the API, and I'd recommend continuing in that directly such that it's fairly easy to replace `rawSystem` with some code that sets the appropriate option.

simonmar:
Ok,sounds good to me too.  Thanks!

snoyberg:
BTW @Mistuke, I have not yet looked at the actual code on this at all, I figured we should hash out this design bit first. I'm assuming I should hold off on an actual code review until the PR is updated to reflect these decisions, do you agree?

Mistuke:
@snoyman yeah,  that would be best since the haskell side is going to
change anyway ☺️ I won't have time to update it till the weekend, but I'll
get it done then.

Mistuke:
Sorry for the delay, I'm almost done, just have to figure out a weird issue with stage-2 when using this version to build gcc.

Mistuke:
Fixed the issue, now ready for review. Thanks!

Mistuke:
Oops, forgot to add the note, I'll do that today.

Mistuke:
Ping

snoyberg:
I haven't had a chance to look into this yet, I'm hoping to this coming
week.

Mistuke:
Cool, no rush, just wasn't sure you knew it was updated :)

snoyberg:
Am aware, keep meaning to review, keep getting dragged away by less
interesting tasks :p

System/Process.hs
```diff
@@ -604,7 +607,14 @@ waitForProcess ph@(ProcessHandle _ delegating_ctlc) = do
         when delegating_ctlc $
           endDelegateControlC e
         return e
-
+    OpenExtHandle _ job iocp -> do
```
snoyberg: Minor note: looks like the `do` is unnecessary

System/Process.hs
```diff
@@ -639,6 +650,10 @@ getProcessExitCode ph@(ProcessHandle _ delegating_ctlc) = do
     Just e | was_open && delegating_ctlc -> endDelegateControlC e
     _                                    -> return ()
   return m_e
+    where getHandle :: ProcessHandle__ -> PHANDLE
+          getHandle (OpenHandle        h) = h
+          getHandle (ClosedHandle      _) = error "getHandle: handle closed."
```
snoyberg: Not a hard requirement, but I'd prefer to avoid the partiality here. I was thinking something like:

```haskell
case getHandle p_ of
  Left e -> return (p_, (Just e, False))
  Right h -> do ...
where
  getHandle (OpenHandle h) = Right h
  getHandle (ClosedHandle e) = Left e
  getHandle (OpenExtHandle h _ _) = Right h
```

System/Process/Common.hs
```diff
@@ -94,13 +95,27 @@ data CreateProcess = CreateProcess{
...
                                            --
                                            --   @since 1.4.0.0
-  child_user :: Maybe UserID               -- ^ Use posix setuid to set child process's user id; does nothing on other platforms.
+  child_user :: Maybe UserID,              -- ^ Use posix setuid to set child process's user id; does nothing on other platforms.
                                            --
                                            --   Default: @Nothing@
                                            --
                                            --   @since 1.4.0.0
+  use_process_jobs :: Bool                 -- ^ On Windows systems this flag indicates that we should wait for the entire process tree
+                                           --   to finish before unblocking. On POSIX system this flag is ignored.
```
snoyberg: s/system/systems

System/Process/Common.hs
```diff
@@ -94,13 +95,27 @@ data CreateProcess = CreateProcess{
...
+                                           --
+                                           --   Default: @False@
+                                           --
+                                           --   @since 1.x.x.x
```
snoyberg: I think we should call this 1.5.0.0

System/Process/Common.hs
```diff
@@ -154,8 +169,14 @@ data StdStream
...
      termination: they all return a 'ProcessHandle' which may be used
      to wait for the process later.
+
+     On Windows a second wait method can be used to block for event
+     completion. This requires two handles. A process job handle and
+     a events handle to monitor.
 -}
-data ProcessHandle__ = OpenHandle PHANDLE | ClosedHandle ExitCode
+data ProcessHandle__ = OpenHandle PHANDLE
+                     | OpenExtHandle PHANDLE PHANDLE PHANDLE
```
snoyberg: Would it make sense to make these fields strict? It does not seem like laziness would ever be a virtue here. (I realize that will make it mismatched with the other constructors, still asking if it's the right thing to do.)

System/Process/Internals.hs
```diff
@@ -24,16 +24,21 @@ module System.Process.Internals (
...
     runGenProcess_, --deprecated
     fdToHandle,
     startDelegateControlC,
     endDelegateControlC,
     stopDelegateControlC,
+    unwrapHandles,
 #ifndef WINDOWS
     pPrPr_disableITimers, c_execvpe,
     ignoreSignal, defaultSignal,
+#else
```
snoyberg: For clarity: could you switch the conditional for `#ifdef WINDOWS`? It's slightly easier to read that way.

System/Process/Internals.hs
```diff
@@ -66,6 +70,18 @@ import System.Process.Posix
 -- * This function takes an extra @String@ argument to be used in creating
 --   error messages.
 --
+-- * 'use_process_jobs' can set in CreateProcess since 1.4.?.? in order to create
```
snoyberg: Probably "can be set", and again let's use 1.5.0.0.

System/Process/Windows.hsc
```diff
@@ -1,4 +1,5 @@
-{-# LANGUAGE CPP #-}
```
snoyberg: I'm going to review this file too, but my review will be much less authoritative. Basically: I'm assuming you know what's going on for Windows better than I do, which is a really safe assumption 😄.

changelog.md
```diff
@@ -9,6 +9,8 @@
 
 * New exposed `withCreateProcess`
 * Derive `Show` and `Eq` for `CreateProcess`, `CmdSpec`, and `StdStream`
+* Add support for monitoring process tree for termination with the parameter `use_process_jobs`
```
snoyberg: This should go into "Unreleased changes"

process.cabal
```diff
@@ -1,5 +1,5 @@
 name:          process
-version:       1.4.3.0
+version:       1.4.3.1
```
snoyberg: As mentioned above, this would be 1.5.0.0

Mistuke:
Seems the appveyor build is having trouble with sourceforge.

snoyberg:
I've restarted the AppVeyor build, which now passed.

System/Process/Common.hs
snoyberg: My question is really: would we ever _want_ the laziness here? I would generally make fields strict if there's no reason to do otherwise, since it excludes a potential source of bugs and confusion. Also, in this case, I think it can (slightly) help performance by allowing unpacking of a primitive type, but that's a minor concern. Either way, this isn't vital.

snoyberg:
Thanks! What are your thoughts on roll-out of a new release for this? Do you want to get more testing of this with GHC before we release?

Mistuke:
It should be OK, the default behavior should be unchanged. Though I would
like to run a x86 build just to be sure. I'll run the full testsuite on
both tonight and report back. Other than that I'm fairly confident. A
similar code has been in use in timeout.hs for a while now. Thanks!

snoyberg:
Sure, thank you!

Mistuke:
Ok, I think we're good. Whenever you want to make the release is fine with me! After you do I can take care of updating the rest of the dependencies to get it in the GHC tree if it's easier for you!

snoyberg:
I've made the release to Hackage. If you could take care of the updates on the GHC side that would be great, much appreciated!

Mistuke:
Alright,  I'll do that then. Thanks!

PR #81, opened 2016-12-12
Don't close already closed pipes 

bgamari:
a9a8e91 added some `close` calls which attempted to close pipes that we already closed previously. This resulted in the error code we read from the failed child being overwritten, resulting in the `process004` test failing.

### DIFF
--- a/cbits/runProcess.c
+++ b/cbits/runProcess.c
@@ -161,7 +161,7 @@ runInteractiveProcess (char *const args[],
         if ((flags & RUN_PROCESS_IN_NEW_GROUP) != 0) {
             setpgid(0, 0);
         }
-        
+
         if ( childGroup) {
             if ( setgid( *childGroup) != 0) {
                 // ERROR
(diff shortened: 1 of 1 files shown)


hvr:
Btw, the changelog entry for 1.4.3 didn't even mention the code change resulting in this regression: http://hackage.haskell.org/package/process-1.4.3.0/changelog

snoyberg:
Thanks @bgamari. Is it reasonable to hold off on a release to Hackage until you've tested this in GHC?

As a side note, it would be really nice to configure Travis/AppVeyor to run the GHC test cases in this repo, I'm just not sure of how to even go about that.

bgamari:
It would be nice if we could make that happen but I to am rather unsure of how to go about it. I can think of a few options,

 * Pulling GHC's testsuite driver into `process` (yuck)
 * Teaching the CI scripts to clone a GHC tree and invoke GHC's testsuite driver with `HC=\`which ghc\``
 * Coming up with a minimal test driver which is loosely compatible with GHC's for inclusion with `process`

PR #91, opened 2017-03-25
Re-export CGid and friends from System.Process.Internals

### LINKED ISSUE #90: Export CGid from System.Process.Internals (especially on Windows)
I found myself wanting to derive some instances for `CreateProcess` in one of my applications recently, only to discover that it wouldn't work on Windows. Why? Because `CreateProcess` uses `UserID` and `GroupID` fields, the definitions of which are platform-dependent:

```haskell
#ifdef WINDOWS
-- Define some missing types for Windows compatibility. Note that these values
-- will never actually be used, as the setuid/setgid system calls are not
-- applicable on Windows. No value of this type will ever exist.
newtype CGid = CGid Word32
  deriving (Show, Eq)
type GroupID = CGid
type UserID = CGid
#else
type PHANDLE = CPid
#endif
```

This Windows-only newtype `CGid` is only available in the internal `System.Process.Common` module, and as a result, I can't make any instances for it (and thus can't derive any instances for `CreateProcess`).

Would you consider reexporting `CGid` from `System.Process.Internals`? That way I could accomplish my original goal, and `process` wouldn't have to make any API promises about the design of `CGid`.

RyanGlScott:
This re-exports `CGid` (which was previously unusable on Windows due to its unavailability) from `System.Process.Internals`. It also re-exports the synonyms `GroupID` and `UserID` so that the Haddocks will have a place to point to on Windows.

Fixes #90.

### DIFF
--- a/System/Process/Common.hs
+++ b/System/Process/Common.hs
@@ -9,6 +9,8 @@ module System.Process.Common
     , ProcRetHandles (..)
     , withFilePathException
     , PHANDLE
+    , GroupID
+    , UserID
     , modifyProcessHandle
     , withProcessHandle
     , fd_stdin
--- a/System/Process/Internals.hs
+++ b/System/Process/Internals.hs
@@ -22,6 +22,13 @@
 module System.Process.Internals (
     ProcessHandle(..), ProcessHandle__(..),
     PHANDLE, closePHANDLE, mkProcessHandle,
+#ifdef WINDOWS
+    CGid(..),
+#else
+    CGid,
+#endif
+    GroupID,
+    UserID,
     modifyProcessHandle, withProcessHandle,
     CreateProcess(..),
     CmdSpec(..), StdStream(..), ProcRetHandles (..),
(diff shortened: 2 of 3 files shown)


RyanGlScott:
I'm not sure what went wrong with the AppVeyor build, but I don't think it was caused by these changes.

snoyberg:
There was a recent change in AppVeyor that broke a bunch of scripts, see http://help.appveyor.com/discussions/problems/6312-curl-command-not-found. Can you rebase this PR against master? Also, can you include a comment in the changelog about the change?

RyanGlScott:
@snoyberg, I've rebased and updated the changelog. The AppVeyor build is still failing for some reason, though.

snoyberg:
I'm not sure what that error means either, let me play with it a bit. I'd like to hold off on merging this until AppVeyor signs off since it's obviously fairly Windows-specific.

snoyberg:
Alright, after a ridiculous amount of pain and suffering I think I've got it fixed. Can you try rebasing against master again?

RyanGlScott:
The Windows build is finally green! Hooray!

snoyberg:
Woohoo!

PR #94, opened 2017-04-22
System.Process.Posix: Hide mb_delegate_ctlc

bgamari:
As it is shadowed by other local bindings in this module.

It turns out this is necessary afterall.


snoyberg:
Thanks, I'm assuming this is necessary for GHC 8.2.

bgamari:
Indeed it is. Thanks!

PR #97, opened 2017-05-22
Export withCreateProcess documentation

jml:
The function is public, but the Haddock documentation isn't being rendered. I think it's a simple matter of removing the `{-` and `-}`, but I haven't tested locally.

Example render of missing documentation:

screen shot 2017-05-22 at 16 38 01

(from http://hackage.haskell.org/package/process-1.6.0.0/docs/System-Process.html)


snoyberg:
Thanks

PR #101, opened 2017-08-03
Allow async exceptions to pierce masked waitForProcess

mitchellwrosen:
Hello!

I believe I found a bug in `waitForProcess`. When run in the `MaskedInterruptible` state, an async exception doesn't actually bring the thread down.

For example, the `typed-process` library exports [`withProcess`](https://hackage.haskell.org/package/typed-process-0.1.0.0/docs/System-Process-Typed.html#v:withProcess) which is similar to [`withCreateProcess`](https://hackage.haskell.org/package/process-1.6.1.0/docs/System-Process.html#v:withCreateProcess), but ends up calling `waitForProcess` in the cleanup action of a `bracket`.

Even though `waitForProcess` is an `interruptible` foreign call, I think what's happening is:

- Async exception enqueued for delivery to thread making `waitForProcess` call
- GHC sees it's in an `interruptible` foreign call, sends `SIGPIPE` ([GHC docs](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/ffi-chap.html#interruptible-foreign-calls))
- [`c_waitForProcess`](https://github.com/haskell/process/blob/c3ab2ccac5972ef23e0727d3c016652e9cda0d1f/System/Process.hs#L592) returns `-1` with errno = `EINTR`
- Upon returning to Haskell, the pending exception is _not_ delivered, because the masking state is `MaskedInterruptible`
- `c_waitForProcess` is retried because it's wrapped by `throwErrnoIfMinus1Retry_`

So our "interruptible" `waitForProcess` is only half-interrupted :)

My fix in this patch is to simply poll for any async exceptions before entering `c_waitForProcess`. [Here](https://github.com/fpco/typed-process/pull/2) is some related discussion I had with @snoyberg.

Thanks!

### DIFF
--- a/test/main.hs
+++ b/test/main.hs
@@ -1,5 +1,5 @@
 import Control.Exception
-import Control.Monad (unless, void)
+import Control.Monad (guard, unless, void)
 import System.Exit
 import System.IO.Error
 import System.Directory (getCurrentDirectory, setCurrentDirectory)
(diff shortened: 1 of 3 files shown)


snoyberg:
This looks good, I feel much better about the usage of `allowInterrupt` here. For the record: I checked and the new test case _does_ fail when `allowInterrupt` is not used.

test/main.hs
```diff
@@ -81,6 +81,17 @@ main = do
       unless (e1 == ExitSuccess && e2 == ExitSuccess)
             $ error "sleep exited with non-zero exit code!"
 
+    do
```
snoyberg: The test implementation is pretty fragile.

* The sleep value is low enough that the process could legitimately finish before the `waitForProcess` begins
* I'd rather not rely on `BlockedIndefinitelyOnMVar`, but instead use `onException` to `putMVar` a value in the case of an exception
* It's worth reporting the `ec` in the `Right` case; an exit code of 0 vs 127 will be highly informative, for instance

mitchellwrosen: Updated - how's that?

snoyberg: Looks good! One last request: can you add a comment to the ChangeLog explaining the change in semantics here?

mitchellwrosen: Done

snoyberg:
Thanks!

Issue #100, opened 2017-07-31
Don't mention typed-process in Cabal description


ivan-m:
In [this commit](https://github.com/haskell/process/commit/ad344892e2b745458c44ffcb079a52a274f683f2), you added a link to the new typed-process library.

Whilst I don't have a problem with that library, I think this is a contentious and surprising link to add:

* The `process` library is officially maintained by the libraries@haskell.org list, and as such I think any promotion of an alternative should be discussed and decided upon within that list.

* Furthermore, by linking to haskell-lang.org from a boot library, this provides it with an extra aura of authority, despite the fact that haskell-lang.org is not an official website for the Haskell language or community.

(Note: I don't think this was done maliciously, just with your "this is a much better library and here's a link to why you should use it" hat on.)

Whilst it may be beneficial to advise users that the typed-process library is safer/better, if it were to be mentioned from the process library (which I still think should be at least discussed on the mailing list) I think a more appropriate way of doing so is along the lines of "You may be interested in ..." and linking to the Hackage (and possibly Stackage) page for that library, which can then itself link to the tutorial.

snoyberg:
I was specifically asked to add such a link to help people out, I was opposed to it initially.

ivan-m:
May I ask by whom and why?

I don't have anything against typed-process, but for something to be semi-officially blessed as a solution in this way is a bit disquieting, especially as it then transitively blesses conduit for streaming; it's also only used by two packages as reported by Stackage, which makes it seem a bit un-battle-tested to be promoted in this way.

snoyberg:
Example: https://www.reddit.com/r/haskell/comments/5w3jdg/the_typedprocess_library_what_to_do_with_it/de7icc4/

Let me say directly: I consider these kinds of discussions just about the most unpleasant kind to exist in package maintainership, and don't enjoy wasting time with them.

erikd:
@snoyberg I really don't want to add to your pain here, but feel I need to weigh in.

I don't think anyone would complain if the text in the commit that @ivan-m references pointed to Hackage rather than the site it does.

Like @jwiegley I am very much in favor of improvements in this direction, especially ones that have been approved by the libraries committee and received more wide spread testing.

Be happy to buy you a beer (or even a scotch) in Oxford if that helps ease any pain involved in this kind of discussion.

simonmar:
Agree with @erikd, let's adjust the link at the very least please.

hvr:
[...]

PR #102, opened 2017-08-16
process.cabal: Use Hackage URL for typed-process

erikd:
Related to #100 .

### DIFF
--- a/process.cabal
+++ b/process.cabal
@@ -16,7 +16,7 @@ description:
     which uses this package internally. It features better binary
     support, easier concurrency, and a more composable API. You can
     read more about it at
-    <https://haskell-lang.org/library/typed-process>.
+    <https://hackage.haskell.org/package/typed-process>.
 
 extra-source-files:
     aclocal.m4


erikd:
Since I'm a member, I will merge this in 48 hours if it has not already been merged and there is no discussion.

snoyberg:
This is not official documentation link for this package. -1

simonmar:
If the haskell-lang URL is the official documentation link, maybe set that URL to be the "Home page" in the .cabal file? Currently it points to the readme on github.

In general I think it's better to use Hackage URLs as the canonical references to packages, since we can find everything else we need to know about a package from that central place (documentation, repo, versions, compatibility, dependencies, etc.), and in a standardised way.

process.cabal
```diff
@@ -16,7 +16,7 @@ description:
     which uses this package internally. It features better binary
     support, easier concurrency, and a more composable API. You can
     read more about it at
-    <https://haskell-lang.org/library/typed-process>.
+    <http://hackage.haskell.org/package/typed-process>.
```
tfausak: Why not HTTPS?

snoyberg:
I didn't realize I had forgotten to update the cabal file, I'll do so a bit later.

You're certainly entitled to believe that Hackage should be the canonical homepage for a package, but I strongly disagree. For my packages, I try to include a thorough tutorial and use that as the homepage. It's why I lobbied so strongly to have prominent placement of READMEs on Hackage. When that was rejected, I started hosting tutorials elsewhere. Additionally, the Haddock generation on Hackage has been unreliable at many points in the past, which is why I always point to Stackage hosted Haddocks.

As I said, you're free to disagree. But I place a high priority on both of these points (tutorials and Haddock reliability), and since I was unable to push any solutions through on Hackage, I no longer treat such URLs as canonical.

hvr:
Indeed, Hackage generation *was* unreliable in *the past*. It works quite well now for the majority of packages that have accurate version bounds. As for the README, I remember that discussion quite well; you requested README support back in 2014, and it was implemented by mid 2015 as a result of your request! READMEs are now rendered in full on the package page quite prominently IMO.

I know you'd rather the `description`s field were deprecated, but that's not sensible for several reasons. In fact, I've recently fixed a couple of things, and now we have almost full Haddock support in the `description`-fields. 

I'm a quite annoyed that everytime things aren't done 100% the way you request it, you find ways to subvert this; a pet peeve of mine is that Stack's template seems to generate packages with the `description` ("Please see README.md") which obviously is not a proper description for your package (and Hackage will show a link to the README at the end of the description *if* the README exists; therefore that phrase is redundant), will will cause poor experience for users (NB: your package won't be properly indexed by Hackage's package search feature and thus hurt its discoverability). See also [this posting of mine](https://mail.haskell.org/pipermail/haskell-cafe/2016-November/125541.html) for more details. But the point being is that defaults in Stack matter, and it's unethical to use it to impose your agenda on Hackage, as a result of that defaults, I've seen *lots* of packages entering the primary Hackage index with a completely useless "see README" description, and often even missing any README; which then fell back to Hackage Trustees to have to point out this quality issue to package maintainers...

snoyberg:
@hvr This completely crosses the line. I have stated what I want. You have decided that I'm wrong and can't have it. My response? I went and implemented it myself on my own, separate project. Now _you're complaining that I implemented it myself_! By your reasoning here, I must agree to your every decision and have no right to dissent. You've now gone further and stated that my disagreement with you is _unethical_. I'm astounded that you believe you have the right to dictate my behavior to this extent.

process.cabal
erikd: Fixed. Thanks!

tfausak:
@hvr: Can you collapse that list, hide it in a `<details>` element, or move it to a Gist? It's annoying to scroll past. 

Also, can you *please* avoid drastically editing your comments after you make them? It makes the conversation hard to follow after the fact. If you have new things to say, like linking to a Stackage bug that didn't exist when you made your original comment, say them in a new comment.

[capture](https://gist.github.com/tfausak/a6afce549961c2aa43ecdecbd17ec44e/revisions?diff=split#diff-d9dde7ab94ed4f4e0216afa621c8893d)

ekmett:
I just realized that this has been merged.

While I'm sympathetic to the arguments that the URL should change, under the [Library_Submissions](https://wiki.haskell.org/Library_submissions) guidelines, Michael is very much within his rights to set it to be wherever he likes:

> Since this has been a difficult point in the past, is bears repeating: *an individual maintainer is fully empowered to make decisions, and should do so!* (emphasis: in the original)

@erikd, merging this commit against the wishes of the package maintainer is very much out of lines with our current documented process. The [appeal process for controversial decisions](https://wiki.haskell.org/Library_submissions#Controversial_decisions) is to ask the Core Libraries Committee to form a consensus, but in the absence of such a judgment or a change in our processes, Michael is free to link wherever he likes.

PR #105, opened 2017-09-25
Bump Win32 constraint up for GHC 8.4

Mistuke:
Hi,

I'd like to bump up the Win32 version for GHC 8.4.

Win32 2.6 contains API breaking changes but those APIs aren't used by process.


snoyberg:
Thanks!

Issue #108, opened 2017-11-11
please add a "getPid :: ProcessHandle -> IO (Maybe Int)"


sternmull:
At the moment there is no way to get the PID of the subprocess. This can be very useful for logging or to use external commands to interact with a process.

This would be the equivalent of https://docs.python.org/3/library/subprocess.html#subprocess.Popen.pid

snoyberg:
I'm not opposed to such an addition, though the caveats that go along with it—specifically that the process may have died since the PID was acquired—need to be spelled out in the docs.

I'm not familiar enough with the Windows side of things to know if this has a natural implementation there, do you have any insight?

Finally, are you interested in writing a PR for this?

EpicOrange:
Seems like this can be closed!

snoyberg:
Thanks for pinging! Closing

PR #109, opened 2017-11-14
Added getPid

sternmull:
This implements my request #108

Tested on Linux and Windows. For Linux i used "stack test". For windows i did a seperate manual test (compared the PID with the info from the Task-Manger). The test/main.hs does not look like it could be run on Windows (for example echo is a builtin of cmd.exe and not a program on its own).

On Windows the PID stays valid as long as the handle is open (see https://blogs.msdn.microsoft.com/oldnewthing/20110107-00/?p=11803).

### DIFF
--- a/System/Process.hs
+++ b/System/Process.hs
@@ -46,6 +46,8 @@ module System.Process (
 
     -- ** Related utilities
     showCommandForUser,
+    Pid,
+    getPid,
 
     -- ** Control-C handling on Unix
     -- $ctlc-handling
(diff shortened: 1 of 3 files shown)


snoyberg:
It looks like the echo bit ran fine on Windows, but for some reason the `terminateProcess` call (implicit in `withCreateProcess`) failed: https://ci.appveyor.com/project/snoyberg/process/build/1.0.186.

I'm also not sure of the motivation behind using an `Integer` value, instead of a `CPid` or similar.

Finally: there's a build failure with GHC 7.2.2: https://travis-ci.org/haskell/process/jobs/302093304.

sternmull:
The problem of withCreateProcess on Windows is unrelated to my changes. I opened #110 for this. The test for getPid now no longer uses withCreateProcess for this reason.

I chose Integer as return type as a lazy solution to have a type that can represent the PID on  all systems (Int32 or possibly larger on POSIX, DWORD on Windows.. which is an unsigned 32bit integer). I knew it was not the optimal solution and did it anyway. I resolved that by adding a Pid type to System.Process that is an alias for the native type that represents the PID.

I saw that your tests run on msys. That explains why i had no success in plain cmd.exe. With a minor adjustment the test now passes on all systems (OSX builds are still pending... but i am confident).

snoyberg:
Thanks for the update and the other bug report. One final request: can you update the changelog and add a `@since` comment to the Haddock?

@simonmar Would you mind having a quick look at this one and making sure I'm not missing some reason why this function would be problematic in the public API?

sternmull:
Updated the comments and changelog.

snoyberg:
Thanks!

Issue #110, opened 2017-11-15
non deterministic exception from withCreateProcess on Windows: "terminateProcess: permission denied (Permission denied)"


sternmull:
On Windows the following program
```.hs
module Main where
import System.Process
import System.IO

main :: IO ()
main = test 1

test i = do
  print i
  withCreateProcess ((shell "echo x") {std_out = CreatePipe}) $ \ _ (Just out) _ p -> do
    hGetChar out -- hGetLine and hGetContents have the same effect. But the problem disappears when this line is removed!
    return ()
  test $ i + 1
```
leads to output like this:
```
1
2
3
4
5
6
7
mytest.EXE: terminateProcess: permission denied (Permission denied)
```
Sometime it happens earlier sometimes it happens later. But as far i can tell it always happens pretty soon. I can not reproduce it on linux.

snoyberg:
Unfortunately I don't know enough about the Win32 API to hazard a reasonable guess as to what's going on here. @Mistuke Could I bother you to look at this bug report and see if anything jumps out at you?

Mistuke:
I don't think this is a bug in `process`, calling `TerminateProcess` will fail with an `Access Denied` when you don't have the rights to terminate the process or when the process is holding on to an object from a higher access level.

In this case what I think is happening is that the blocking calls to `hGetChar` end up creating `IRPs` which the kernel can't cancel for some reason. Reason may be buggy I/O driver, or a bug in the GHC I/O manager which would not be surprising. We might be creating requests which are issued not being cancellable because the I/O manager on Windows currently does not use native I/O facilities like IOCP. Instead relies on the pseudo posix interfaces which are largely un-maintained and deprecated.

That's why the issue goes away when you don't use an `I/O` function.

You can use a kernel mode debugger (like `livekd`) to inspect the `IRPs` when the process refused to die. I'll try to reproduce it later and see if I can find where it's going wrong.

snoyberg:
Wow, thanks for the thorough and quick response :)

sternmull:
See #111

thomasjm:
Hi all -- I'm seeing this issue also, is #111 going to land?

sternmull:
I still hope so. There was some misunderstanding in the discussion for #111 but in the end we agreed the fix is good.

PR #111, opened 2017-11-16
#110, On Windows ignore ERROR_ACCESS_DENIED for TerminateProcess() if…

sternmull:
It seems this can be attributed to strange behavior of the Windows API. libuv has a workaround for this: https://github.com/libuv/libuv/blob/master/src/win/process.c#L1172

It is not perfect (because it does work only for processes that do not exit with 259). But to my current knowledge this is a platform specific limitation we have to accept.


Mistuke:
Libuv doesn't swallow the error, it returns `UV_ESRCH` which is no such process. So these two hammers are not the same. Libuv correctly propagates the fact that the process is still alive. This just pretends it's not, and that's critical for correctness because the process can still return STILL_ALIVE for a number of reasons.

sternmull:
The change that i propose only ignores `ERROR_ACCESS_DENIED` of `TerminateProcess()` if the process did die (and did not exit with `STILL_ACTIVE`). In this case we can be sure the process already has terminated. So this change only makes `terminateProcess` succeed also when the process is already gone (even if `TerminateProcess()` fails to succeed for that reason).

Mistuke:
Ah that's true. I missed the !, still this is not solving your original
problem. It's just hiding it.

sternmull:
It solves it... except for the situation where the subprocess exits with 259 (`STILL_ACTIVE`) where the `ERROR_ACCESS_DENIED` is propagated just like before. So for example my test case from #110 now runs without errors and correctly terminates its subprocesses. See it more as an improvement than a complete solution.

I think we have to accept that there is a chain of obstacles (`TerminateProcess()` sometimes failing for mysterious undocumented reasons, `GetExitCodeProcess()` being helpful to work around that... except for one specific exit code) and have to make the best out of it.

Mistuke:
How does it solve it. Your process wasn't dead. It was just blocked.

sternmull:
What makes you think that? The code path that i added only succeeds if the process actually terminated. So you can be sure the process was dead when `terminateProcess` succeeded.

Mistuke:
Oh I see, I misinterpreted your example, the hGetChar is in the caller not the child. fair enough.

sternmull:
@snoyberg This is waiting for quite a while now. Could you merge it?

snoyberg:
@Mistuke Are you OK with me merging and including this in the next release?

Mistuke:
@snoyberg yes no objections from me. Thanks.

snoyberg:
@sternmull Just one last request then: would you be able to add a short description of the change into the changelog? Thanks!

sternmull:
@snoyberg sure. Rebased the PR on master and updated changelog.

PR #112, opened 2017-12-26
Process: fix flaky test.

Mistuke:
Calling `sh` and `sleep` can be quite flacky on Windows due to the msys2 runtime. 

It can result in random failures such as:

```
+      6 [main] sh 22832 child_copy: cygheap read copy failed, 0x1802F6408..0x1802FF298, done 0, windows pid 22832, Win32 error 299
+  16044 [main] sh 22832 C:/msys64/usr/bin/sh: *** fatal error in forked process - ccalloc would have returned NULL
```

So about that by using a Haskell program that does the same action. This makes the test stable again under Windows.

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -10,3 +10,6 @@ include $(TOP)/mk/test.mk
 process007_fd:
 	'$(TEST_HC)' -optc='-Wall' -no-hs-main -no-auto-link-packages process007_fd.c -o process007_fd
 
+.PHONY: T3994app
+T3994app:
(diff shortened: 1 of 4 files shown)


Mistuke:
Wow my sentences made no sense haha. Thanks @snoyberg

snoyberg:
No problem. I should also set up AppVeyor on this repo :)

snoyberg:
Oh, never mind, AppVeyor is already set up :)

Mistuke:
Yeah, I think the problem is you can't run the GHC bounded tests without GHC. We have an open item to move the testsuite driver to it's own repo. This would allow you to run these tests without a full build of GHC.

PR #113, opened 2017-12-27
Process: fix linux test build.

Mistuke:
Previous PR broke the test under Linux and MacOS due to the new binary not being on the `PATH`.

This version seems to work on Linux and Windows. Excuse the noise :(

### DIFF
--- a/tests/T3994.hs
+++ b/tests/T3994.hs
@@ -5,7 +5,7 @@ import System.IO
 import System.Process
 
 main :: IO ()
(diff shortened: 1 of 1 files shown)


snoyberg:
Just to confirm: this is fixing a test failure in GHC's test suite run, correct? Because the previous PR passed on Travis for both Linux and OS X.

Mistuke:
Yeah, I don't believe the Travis or AppVeyor run these tests, they seem to run the ones in the "test" folder but not in the "tests" folder, which require the GHC testsuite driver.

PR #123, opened 2018-04-03
Close pipes on failure

### LINKED ISSUE #122: Leaking file descriptors
I got a file descriptor leak on production. Traced to the `process` library. Basically, creating commutication pipe may fail, and in that case we leak already allocated pipes for stdin, stdout and strerr. The relevant code is [here](https://github.com/haskell/process/blob/d3d637d2b01748d9807f51f4bb475573a55b24e5/cbits/runProcess.c#L113).

Yuras:
When we fail to create communication pipe, we should cleanup resources
before exit. E.g. we should close pipes for stdin, stdout and stderr if
they where created. Fixes #122.


Yuras:
Done. Should I also update package version in cabal file?

PR #125, opened 2018-04-16
Bump upper bound on unix to < 2.9

RyanGlScott:
See https://ghc.haskell.org/trac/ghc/ticket/15042.


RyanGlScott:
Ack, that commit message is just plain wrong. Let me try again.

RyanGlScott:
Take two: #126.

PR #126, opened 2018-04-16
Bump upper bound on unix to < 2.9

RyanGlScott:
See https://ghc.haskell.org/trac/ghc/ticket/15042.


snoyberg:
Thanks! BTW, feel free to push such changes to master in the future if you'd like.

PR #130, opened 2018-05-18
expose cleanupProcess

srid:
`cleanupProcess` needs to be exported in order to be able to write a version of `withCreateProcess` that works with MonadIO.

### DIFF
--- a/System/Process.hs
+++ b/System/Process.hs
@@ -43,6 +43,7 @@ module System.Process (
     readCreateProcessWithExitCode,
     readProcessWithExitCode,
     withCreateProcess,
+    cleanupProcess,
 
     -- ** Related utilities
     showCommandForUser,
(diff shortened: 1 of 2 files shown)


System/Process.hs
```diff
@@ -245,7 +246,10 @@ withCreateProcess_ fun c action =
     C.bracketOnError (createProcess_ fun c) cleanupProcess
                      (\(m_in, m_out, m_err, ph) -> action m_in m_out m_err ph)
 
-
+-- | Cleans up the process.
+-- 
+-- This function is meant to be invoked from any application level cleanup 
+-- handler. It terminates the process, and closes any 'CreatePipe' 'handle's.
```
snoyberg: LGTM. Can you add a `@since` comment with a new minor version and update the `ChangeLog` with an entry about this new function?

srid: @snoyberg Done.

PR #133, opened 2018-06-12
process00{1,2} tests: just assume `ls` is in the PATH on non-Windows systems

alpmestan:
This should work on all the systems I've ever laid my hands on (not many, admittedly) and of particular interest to me and several other GHC/core libraries contributors who run GHC's testsuite, with this patch those tests now pass on NixOS.

``` sh
$ TEST="process001 process002" ./validate --fast --testsuite-only
[...]
==== STAGE 2 TESTS ==== 

SUMMARY for test run started at Tue Jun 12 15:51:02 2018 CEST
 0:00:01 spent to go through
       2 total tests, which gave rise to
      14 test cases, of which
      12 were skipped

       0 had missing libraries
       2 expected passes
       0 expected failures

       0 caused framework failures
       0 caused framework warnings
       0 unexpected passes
       0 unexpected failures
       0 unexpected stat failures
```

There, `ls` is not under `/bin` but somewhere in the nix store and brought into the PATH through symlinks in a "current environment" directory which itself _is_ in the PATH.

### DIFF
--- a/tests/process001.hs
+++ b/tests/process001.hs
@@ -2,15 +2,9 @@
 import System.IO
 import System.Process
 
-#ifdef mingw32_HOST_OS
-cmd = "ls"
-#else
-cmd = "/bin/ls"
-#endif
-
 test = do
   h <- openFile "process001.out" WriteMode
-  ph <- runProcess cmd [] Nothing Nothing Nothing (Just h) Nothing
+  ph <- runProcess "ls" [] Nothing Nothing Nothing (Just h) Nothing
   waitForProcess ph
 
 main = test >> test >> return ()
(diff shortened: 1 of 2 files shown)


snoyberg:
@bgamari Since this affects the GHC test suite and not the process test suite, would you mind giving it a review? I don't see any issues myself, but I'm not very familiar with GHC's test suite setup.

@Mistuke The changes here touch on Windows-specific code, but seem to leave the Windows code path identical to what it was before, so I don't believe it should affect Win32 at all.

int-index:
@snoyberg Ben has expressed his approval of this change here: https://github.com/alpmestan/ghc.nix/issues/10#issuecomment-396587255

snoyberg:
Thanks @int-index

PR #135, opened 2018-06-15
Add bounds on base in test-suite stanza

bgamari:
GHC's build system runs `ghc-cabal check` on the cabal files for all of its
core libraries. One of the checks that this performs is to verify that base
has bounds. IMHO it's a bit silly that this is necessary even when base
is otherwise constrained by the library stanza so I've left the bounds
quite loose.

### DIFF
--- a/process.cabal
+++ b/process.cabal
@@ -85,7 +85,7 @@ test-suite test
   hs-source-dirs: test
   main-is: main.hs
   type: exitcode-stdio-1.0
-  build-depends: base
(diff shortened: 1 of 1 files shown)


snoyberg:
Would you mind adding an explanation of why the bound is present? The incongruity with the other dependencies may lead someone (like myself) to try to clean this up in the future without a comment.

bgamari:
Sure.

bgamari:
Superceded by #136.

Issue #50, opened 2015-12-04
NoStream should reopen on /dev/null / NUL:


geekosaur:
Closing the standard file descriptors (stdin, stdout, stderr) when starting a new process is incorrect, and can lead to buggy behavior in the child (for example, GNU libc++ will abort() if stdin is closed) or potentially information leaks if the forked process itself opens a file descriptor it does not intend a child to inherit and it lands on the closed standard fd. The correct behavior is to open /dev/null on the fd.

I'm working on a patch, but am wondering what to do if the reopen doesn't work correctly. Currently it behaves as if the exec() failed due to the security implications of passing a closed standard filehandle, since there is no defined fork*Failed value for these.

I also note the dup2-s for pipes are not checked for failure and may lead to similar issues.

GAumala:
I was testing this library running bash scripts with simple commands like `ls` and `echo` and they fail when I set `std_out` to `NoStream`, but succeed when I set it to `Inherit` or `CreatePipe`. It's super puzzling because without a stream it's impossible to debug the script.

snoyberg:
Only noticing this issue now, I think it predates my role as maintainer here.

I think this should be addressed via docs. There is a semantic difference between closing streams and pointing at /dev/null, and we do not want to silently change that behavior. There _are_ times (though perhaps rare) where the former is exactly what's desired.

GAumala:
Yeah I agree, updating documentation would be good enough. I could try working on a PR. If I wanted to run a new process, completely ignoring stderr and stdout, `CreatePipe` is probably the best choice right?

snoyberg:
Not on its own. Unless you drain the created pipes in the parent process, the OS buffer will ultimately fill up and writes to `stdout`/`stderr` in the child will block indefinitely. You need to either drain those pipes (likely in a separate thread), write to a temp file and delete it, or write to something like `/dev/null`.

snoyberg:
Addressed by doc update in #138, thanks @GAumala!

PR #138, opened 2018-09-20
Improve documentation for NoStream

GAumala:
As discussed on #50 improve the documentation for `NoStream`


snoyberg:
That's not quite what I would have expected. I would have liked to see an explanation that:

* It provides closed file descriptor to the child process
* The child process will likely error out if it tries to write to them
* If you need "ignoring" behavior, you'll need to create a pipe and ignore the data from the `Handle` yourself.

PR #141, opened 2018-12-12
Fix a -Wincomplete-uni-patterns warning

ckoparkar:
This warning will soon be enabled by -Wall (See Trac #15656).

### DIFF
--- a/System/Process.hs
+++ b/System/Process.hs
@@ -454,25 +454,30 @@ readCreateProcess cp input = do
                     std_out = CreatePipe
                   }
     (ex, output) <- withCreateProcess_ "readCreateProcess" cp_opts $
-      \(Just inh) (Just outh) _ ph -> do
+      \mb_inh mb_outh _ ph ->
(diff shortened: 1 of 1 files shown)


ckoparkar:
The incomplete patterns:

1. [`readCreateProcess`](https://github.com/haskell/process/blob/master/System/Process.hs#L457)
2. [`readCreateProcessWithExitCode`](https://github.com/haskell/process/blob/master/System/Process.hs#L526)

snoyberg:
Yes, I'd much prefer that to turning off this warning. Would you be able to update the PR?

ckoparkar:
@snoyberg Yes, I realize that's what I should have done, instead of suppressing the warning. Should I use plain `error`, or would that break the semantics of other things which use `IOException`s ?

snoyberg:
Using error should be fine.

ckoparkar:
@snoyberg fixed it to  use `error`.

snoyberg:
Thanks!

PR #148, opened 2019-07-01
Call initgroups before setuid

### LINKED ISSUE #149: Process fails to call initgroups before setuid
On POSIX platforms we use `setuid` and `setgid` to change the child process's uid and gid. This is most often used to drop privileges. However, `setgid` merely changes a processes' primary group; it does not drop the other groups which the process may also be a member of. To do this we must call `initgroups`.

The fact that we fail to call `initgroups` means that we may not be dropping as many privileges as the user expects.

bgamari:
This fixes a potential privilege escalation issue (or, more precisely, privileges not being dropped when this was the user's intent) where the groups of the spawning process's user would be incorrectly retained due to a missing call to `initgroups`.

This also fixes an incorrect case fallthrough in the error message lookup logic.

Fixes #149.

### DIFF
--- a/cbits/runProcess.c
+++ b/cbits/runProcess.c
@@ -329,6 +329,7 @@ runInteractiveProcess (char *const args[],
             break;
         case forkSetuidFailed:
             *failed_doing = "runInteractiveProcess: setuid";
+            break;
         default:
             *failed_doing = "runInteractiveProcess: unknown";
             break;
--- a/cbits/runProcess.c
+++ b/cbits/runProcess.c
@@ -33,6 +33,10 @@ static long max_fd = 0;
 extern void blockUserSignals(void);
 extern void unblockUserSignals(void);
 
+// These are arbitrarily chosen -- JP
+#define forkSetgidFailed 124
+#define forkSetuidFailed 125
+
 // See #1593.  The convention for the exit code when
 // exec() fails seems to be 127 (gleened from C's
 // system()), but there's no equivalent convention for
@@ -40,9 +44,8 @@ extern void unblockUserSignals(void);
 #define forkChdirFailed 126
 #define forkExecFailed  127
 
-// These are arbitrarily chosen -- JP
-#define forkSetgidFailed 124
-#define forkSetuidFailed 125
+#define forkGetpwuidFailed 128
+#define forkInitgroupsFailed 129
 
 __attribute__((__noreturn__))
 static void childFailed(int pipe, int failCode) {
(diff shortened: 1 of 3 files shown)


bgamari:
This was originally reported at https://github.com/ndmitchell/ghcid/issues/261.

snoyberg:
I'm surprised that appveyor is failing on this. I'll try rebasing on `master` and see if that results in something else.

include/runProcess.h
```diff
@@ -21,6 +21,8 @@
 
 #include <unistd.h>
 #include <sys/types.h>
+#include <pwd.h>
```
snoyberg: I think this should be conditional on the OS _not_ being Windows, IIUC that's where the AppVeyor failure is coming from.

cbits/runProcess.c
```diff
@@ -182,6 +185,22 @@ runInteractiveProcess (char *const args[],
...
+            // However, to do this we must know the username of the user we are
+            // switching to.
+            struct passwd pw;
+            struct passwd *res = NULL;
+            int buf_len = sysconf(_SC_GETPW_R_SIZE_MAX);
+            char *buf = malloc(buf_len);
+            gid_t suppl_gid = childGroup ? *childGroup : getgid();
+            if ( getpwuid_r(*childUser, &pw, buf, buf_len, &res) != 0) {
+                childFailed(forkCommunicationFds[1], forkGetpwuidFailed);
+            } else if (res == NULL) {
```
snoyberg: The mixing of `else if` and `if` makes the code confusing, at least IMO. `childFailed` will cause the program to exit regardless, so the `else` isn't strictly necessary, but it arguably helps with readability. I don't care too much one direction or the other, but I'd prefer to see the code be consistent.

If there's a reason for using `else if` in just one of the checks, please clarify, ideally with an inline comment.

juhp:
I have tested Ben's patch with ghc-8.6.5, process:initgroups, and cabal-install-2.2 and can confirm that with the patched process, ghcid-0.7.5 has `initgroups`. Thanks @bgamari for this.

```
[ f31container | ~/process:initgroups ] $ readelf -a ~/.cabal/bin/ghcid|grep initgroups
0000008ba160  005700000007 R_X86_64_JUMP_SLO 0000000000000000 initgroups@GLIBC_2.2.5 + 0
    87: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND initgroups@GLIBC_2.2.5 (4)
```

cbits/runProcess.c
bgamari: Indeed looking back at this I can't imagine what I was thinking. This `else if` looks just plain wrong.

bgamari: Ahh, no, it's not wrong but indeed the `else` isn't necessary. Fixed.

include/runProcess.h
bgamari: Strange how `unistd.h` is available on Windows yet `pwd.h` is not. Regardless, this is now properly guarded.

juhp:
Seems to build okay on Windows now :+1: 
(Not quite sure why lts-11 keep failing but does seems related?)

bgamari:
@snoyberg, done.

PR #151, opened 2019-09-06
Add comment about slow close_fds

typetetris:
Just a quick additional comment, to make people aware, that close_fds is potentially expensive. I was hit by this in ndmitchell/shake#722.


snoyberg:
Thanks!

PR #154, opened 2019-09-16
Export runInteractiveProcess_lock on Posix systems

### LINKED ISSUE #15: Forking can interact with runtime to cause livelock (100% cpu usage) or freezing
If `spawnWithPty` is called at an unlucky time (i.e. when other system calls like printing to stderr and launching other processes are happening) then some kind of bad interaction with the runtime can get triggered, resulting in either full CPU livelock (stuck at 100% across all CPUs) or freezing of the Haskell process.

A nice repro (courtesy of u/Noughtmare on Reddit) is below. You can also find my original, slightly more complicated version as a full repo [here](https://github.com/thomasjm/haskell-livelock-repro).

```haskell
module Main where

import Control.Concurrent
import Control.Monad
import System.Posix.Pty
import System.Process

main :: IO ()
main = do
  forkIO $ replicateM_ 1000 $ putStrLn "This is a test"

  -- Repeat this 20 times
  replicateM_ 20 $ do
    forkIO $ void $ spawnWithPty Nothing True "sleep" ["99999999"] (80, 24)
    forkIO $ void $ readCreateProcess (proc "echo" ["hi"]) ""

  -- All CPUs should now be stuck at 100%
  threadDelay 60000000
```

I'm filing this issue after a Reddit discussion which determined that `posix-pty` most likely needs to be more careful about how it forks. You can see the discussion here: https://www.reddit.com/r/haskell/comments/d4h7ty/why_does_this_code_livelock_all_my_cpus_at_100/

thomasjm:
This PR is to help fix https://github.com/merijn/posix-pty/issues/15. 

Summary: on Posix systems, `System.Process` must take certain precautions before using the `fork` system call. These include altering global state by calling `blockUserSignals` and `stopTimer()` before forking; see https://github.com/haskell/process/blob/master/cbits/runProcess.c#L137.

However, `System.Process` is not the only library that might need to use fork. `System.Posix.Pty` is another such library -- it uses the `forkpty` call (which internally does `fork`) to create a process attached to a pseudo-terminal. Thus System.Posix.Pty also needs to take these precautions.

As a result, we need to ensure mutual exclusion between `System.Posix.Pty` fork calls and `System.Process` fork calls, or else they stomp on each other's changes to global state. The only way I can see to do this is to use the same lock, so this PR exports it. I made a PR on `posix-pty` to leverage it here: https://github.com/merijn/posix-pty/pull/16.

Of course, this feels like a bit of a band-aid since it only ensures that interleaving these two libraries is safe. Other libraries that might want to `fork` will still run into problems interleaving with these libraries. It would be nice to get some GHC expert attention on this but maybe that's an issue for another tracker.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,6 +1,6 @@
 # Changelog for [`process` package](http://hackage.haskell.org/package/process)
 
-## Unreleased changes
+## 1.6.5.2 *September 2019*
 
 * Fix a potential privilege escalation issue (or, more precisely, privileges
   not being dropped when this was the user's intent) where the groups of the
@@ -9,6 +9,8 @@
 * Bug fix: Prevent stripping undecodable bytes from environment variables
   when in a non-unicode locale.
   [#152](https://github.com/haskell/process/issues/152)
+* Expose `runInteractiveProcess_lock` in `System.Process.Internals`
+  [#154](https://github.com/haskell/process/pull/154)
 
 ## 1.6.5.1 *June 2019*
 
--- a/changelog.md
+++ b/changelog.md
@@ -1,6 +1,6 @@
 # Changelog for [`process` package](http://hackage.haskell.org/package/process)
 
-## 1.6.5.2 *September 2019*
+## 1.6.6.0 *September 2019*
 
 * Fix a potential privilege escalation issue (or, more precisely, privileges
   not being dropped when this was the user's intent) where the groups of the
--- a/System/Process/Internals.hs
+++ b/System/Process/Internals.hs
@@ -46,6 +46,7 @@ module System.Process.Internals (
 #else
     pPrPr_disableITimers, c_execvpe,
     ignoreSignal, defaultSignal,
+    runInteractiveProcess_lock,
 #endif
     withFilePathException, withCEnvironment,
     translate,
--- a/System/Process/Posix.hs
+++ b/System/Process/Posix.hs
@@ -17,6 +17,7 @@ module System.Process.Posix
     , createPipeInternal
     , createPipeInternalFd
     , interruptProcessGroupOfInternal
+    , runInteractiveProcess_lock
     ) where
 
 import Control.Concurrent
--- a/System/Process/Posix.hs
+++ b/System/Process/Posix.hs
@@ -179,6 +179,8 @@ createProcess_Internal fun
 -- (and also need to make the same global state changes) can protect their changes
 -- with the same lock.
 -- See https://github.com/haskell/process/pull/154.
+--
+-- @since 1.6.6.0
 runInteractiveProcess_lock :: MVar ()
 runInteractiveProcess_lock = unsafePerformIO $ newMVar ()
 
--- a/System/Process/Posix.hs
+++ b/System/Process/Posix.hs
@@ -139,10 +139,7 @@ createProcess_Internal fun
      when mb_delegate_ctlc
        startDelegateControlC
 
-     -- runInteractiveProcess() blocks signals around the fork().
-     -- Since blocking/unblocking of signals is a global state
-     -- operation, we better ensure mutual exclusion of calls to
-     -- runInteractiveProcess().
+     -- See the comment on runInteractiveProcess_lock
      proc_handle <- withMVar runInteractiveProcess_lock $ \_ ->
                          c_runInteractiveProcess pargs pWorkDir pEnv
                                 fdin fdout fderr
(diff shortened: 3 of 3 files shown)


snoyberg:
I don't feel strongly either way, but overall I'm concerned that there may be undocumented invariants, and exposing this locking primitive will end up encouraging misuse.

thomasjm:
Yes, agreed this is not pretty but it seems to be forced by the design of the RTS. If there were some facility provided by the language itself to ensure that an FFI fork call is safe, then libraries wouldn't have to come up with their own locking schemes.

As it stands, other libraries can't safely fork if that operation might get interleaved with a `System.Process` call, and I think we should fix that ASAP. For me it produced a quite nasty intermittent livelock bug. Surfacing this flaw might help other people avoid the same while a more principled fix is investigated.

It seems like a reasonable compromise to put it in `System.Process.Internal`, no? That kind of communicates the sketchiness of the fix and allows us to remove it someday later.

snoyberg:
Fair enough. In that case, can you add a changelog entry, and some Haddocks on the `lock` explaining what it is and why it's exposed, possibly with a link to this PR?

thomasjm:
Done!

changelog.md
```diff
@@ -1,6 +1,6 @@
 # Changelog for [`process` package](http://hackage.haskell.org/package/process)
 
-## Unreleased changes
+## 1.6.5.2 *September 2019*
```
snoyberg: Version bump should be to 1.6.6. Also, please include a `@since` comment on the new Haddocks.

thomasjm: Done

snoyberg (approved):
Thank you!

thomasjm:
Np! Any chance I could get you to cut a Hackage release @snoyberg ?

snoyberg:
Done!

Merivuokko:
I am writing a library which runs processes kinda like `process` does.
The reason I don't use `process` is that it does not support `OsPath`s.
I need to support whatever filenames the OS supports.

I wanted to see, if `process` does something special on Posix systems and I was surprised in a very unpleasant way when I found this issue.

I believe this lock belongs to the `unix` package as it exports fork.
I don't want to depend on `process` just for this lock.
I also have no way of knowing whether my library might be used in a program which also uses `process`.

If `unix` cannot/does not want to own this lock, maybe a new package could be created, containing only the lock, maybe together with a version of `fork` which takes care of the safety precautions.

Was this issue ever discussed with GHC developers?

tomjaguarpaw:
Hello @Merivuokko, I'm afraid I don't entirely understand what you're asking. Are you saying there is some functionality which ought to be in [`unix`](https://hackage.haskell.org/package/unix)? If so, have you made this suggestion to the `unix` maintainers, perhaps by filing a [feature request](https://github.com/haskell/unix/issues/new?template=feature_request.md)?

Or is there something in particular you are asking from the `process` maintainers?

Merivuokko:
Hi,

I'm sorry of being unclear.

Basically this issue discusses problems of using the fork system call on POSIX
platforms with GHC runtime. A race condition may be triggered if two threads
fork simultaneously. In my opinion fork is a very commonly needed primitive,
although in Haskell we often use libraries (like process) which abstract away
forking. For this reason, there should be a commonly accepted interface for
safe forking, which could be shared by all libraries/applications that need to
fork.

Currently `process` is the package that provides the safety mechanism (the
lock discussed in this issue). `process`, however, is quite a high-level
library, so does this lock actually belong to process?

My intention was to ask, what would be the right way to attempt to solve this
issue. It seems that GHC or unix packages would be good candidates for
providing a shared and safe way for a Haskell process to fork.

So I'm just posting questions:

- What do you process maintainers think, where should this lock be, or should
  there even be some other way of resolving the race condition?

- This issue mentions a need for discussing this with a GHC maintainer? Did
  this discusssion ever happen?

I would of course be very happy, if `process` provided an interface for
OsPaths, and that would resolve my original problem. However, providing this
interface does not resolve more general need for safe forking.

Hopefully this makes more sense now. Thank you for your patience!

-- 
Aura

tomjaguarpaw:
> It seems that GHC or unix packages would be good candidates for providing a shared and safe way for a Haskell process to fork

This seems like a reasonable point of view to me. I think you're unlikely to get a good discussion on the `process` issue tracker about whether the locking primitive belongs to the `unix` package, so I suggest you ask on the `unix` issue tracker in the first instance.

thomasjm:
Putting it in `unix` wouldn't make sense for Windows, right? I'm not sure about the situation there but Windows may need a lock around process creation.

PR #159, opened 2019-11-05
Windows: Check if all processes really exited and released resources

Mistuke:
The PR makes sure we're only listening to completion events for the `key` that we're interested in. This isn't much of an issue now but the new I/O manager for Windows will exclusively use I/O completion ports. Which means this method may mistakenly handle events it shouldn't have.

Secondly it fixes a rare but annoying condition. When you receive a notification on the port that a process has terminated this does not mean that the resources of the process has been flushed.

Previously we were immediately unregister the process once we get the notification. However things like `hsc2hs` which rely on the child process having written a file then may get into a race condition. Control may return to the caller before the file has actually been written.

Windows deals with this by returning an exit code `STILL_ACTIVE` which is to be interpreted as `"process has exited but has not yet been cleaned up"`.  If we get such an event we then block and wait for the OS to signal the handle and re-set the exit code.

This fixes https://gitlab.haskell.org/ghc/ghc/issues/17108 and https://gitlab.haskell.org/ghc/ghc/issues/17249


Mistuke:
Done :)

PR #172, opened 2020-03-14
Avoid redundant pattern match warning on Windows

bgamari:
GHC is able to see that the the `ClosedHandle` alternative is unreachable due to the containing `case` analysis.


bgamari:
I hope you don't mind if I merge this in the interest of moving the GHC release process along.

PR #176, opened 2020-05-11
Windows: Fix buffer size of QueryInformationJobObject request

bgamari:
As noted in GHC #17926, the QueryInformationJobObject system call
apparently relies on the buffer size in addition to the
NumberOfAssignedProcesses field to determine how many results it should
return. For this reason it's important that we pass the true size of the
buffer to the call.


domenkozar:
Any chance to have a bugfix release?

snoyberg:
Released

PR #177, opened 2020-05-20
Add support for WinIO to process.

Mistuke:
This adds support for the new Asynchronous I/O manager in GHC for Windows[1] to process.

This is required because process uses internal GHC APIs.

Do not merge yet, however goal for this is for GHC 8.12. It is ready for review.

[1] https://gitlab.haskell.org/ghc/ghc/-/merge_requests/1224


process.cabal
```diff
@@ -72,6 +74,7 @@ library
         processFlags.h
 
     ghc-options: -Wall
+    cc-options: -O0 -g3
```
Mistuke, line 77: oops need to remove this.

include/runProcess.h
```diff
@@ -88,6 +88,19 @@ extern ProcHandle runInteractiveProcess( wchar_t *cmd,
                                          bool useJobObject,
                                          HANDLE *hJob );
 
+extern ProcHandle runInteractiveProcessHANDLE ( wchar_t *cmd,
```
Mistuke, line 91: Needs an ifdef guard.

System/Process/Windows.hsc
```diff
@@ -143,17 +245,25 @@ createProcess_Internal fun CreateProcess{ cmdspec = cmdsp,
...
+     Mgr.associateHandle' =<< peek pfdStdInput
+     Mgr.associateHandle' =<< peek pfdStdOutput
+     Mgr.associateHandle' =<< peek pfdStdError
+
+     -- Create the haskell mode handles as files.
+     hndStdInput  <- mbPipeHANDLE mb_stdin  pfdStdInput  WriteMode
+     hndStdOutput <- mbPipeHANDLE mb_stdout pfdStdOutput ReadMode
+     hndStdError  <- mbPipeHANDLE mb_stderr pfdStdError  ReadMode
+
+     print ":: createProcess_Internal_winio - finished"
```
Mistuke, line 261: remove debug code.

cbits/win32/runProcess.c
```diff
@@ -0,0 +1,573 @@
...
+
+/*
+ * Function: mkNamedPipe
+ *
+ * Purpose:  create an named pipe with read and write ends being
+ *           optionally (non-)inheritable. Named pipes can ve read
+ *           asynchronously while anonymous pipes require blocking calls.
+ */
+static BOOL
+mkNamedPipe (HANDLE* pHandleIn, BOOL isInheritableIn,
```
Mistuke, line 91: This is new.

cbits/win32/runProcess.c
```diff
@@ -0,0 +1,573 @@
...
+    {
+        return hJob;
+    }
+
+    maperrno();
+    return NULL;
+}
+
+static inline bool
+setStdHandleInfo (LPHANDLE destination, HANDLE _stdhandle,
```
Mistuke, line 186: This is also new after a refactoring

cbits/win32/runProcess.c
```diff
@@ -0,0 +1,573 @@
...
+                                 DUPLICATE_SAME_ACCESS);
+        if (!status) return false;
+        *destination = *tmpHandle;
+    }
+
+    return true;
+}
+
+static ProcHandle
+runInteractiveProcessWrapper (
```
Mistuke, line 227: Ditto this

System/Process/Common.hs
```diff
@@ -258,3 +269,18 @@ pfdToHandle pfd mode = do
   let enc = localeEncoding
 #endif
   mkHandleFromFD fD' fd_type filepath mode False {-is_socket-} (Just enc)
+
+#if defined(__IO_MANAGER_WINIO__)
+mbHANDLE :: HANDLE -> StdStream -> IO HANDLE
+mbHANDLE _std CreatePipe      = return $ intPtrToPtr (-1)
+mbHANDLE  std Inherit         = return std
+mbHANDLE _std NoStream        = return $ intPtrToPtr (-2)
```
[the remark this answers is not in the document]
Mistuke: correct, and also `-1` has a different meaning. But I think it's reasonably safe (famous last words). But ideally `process` should use a structure here. but it's more invasive change.

cbits/win32/runProcess.c
```diff
@@ -0,0 +1,573 @@
...
+ */
+static BOOL
+mkAnonPipe (HANDLE* pHandleIn, BOOL isInheritableIn,
+            HANDLE* pHandleOut, BOOL isInheritableOut)
+{
+    HANDLE hTemporaryIn  = NULL;
+    HANDLE hTemporaryOut = NULL;
+
+    /* Create the anon pipe with both ends inheritable */
+    if (!CreatePipe(&hTemporaryIn, &hTemporaryOut, NULL, 0))
```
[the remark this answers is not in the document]
Mistuke: I already do. see the call below this one :) this path is only used for the old I/O system.

Mistuke:
Arg... looks like `createPipe` and `createPipeFd` are exposed. So will have to change those too. `createPipeFd` will just error out. I wonder though... what's the use case of `createPipeFd`?

Mistuke:
Looks like I have to drop the test using the new manager explicitly.
But this needs a small fix for older GHCs and should be done.

I will push changes for `createPipe` in a different `pr`. Any comments here @AndreasPK @snoyberg ?

cbits/win32/runProcess.c
```diff
@@ -0,0 +1,611 @@
...
+                     GENERIC_WRITE,
+                     FILE_SHARE_WRITE,
+                     &secAttr,
+                     OPEN_EXISTING,
+                     FILE_FLAG_OVERLAPPED,
+                     NULL);
+    if (hTemporaryOut == INVALID_HANDLE_VALUE)
+      goto fail;
+
+    /* Set some optimization flags to makethe I/O manager operate more
```
bgamari, line 152: Missing space here.

Thanks for all of these comments, by the way. This makes it much easier to follow.

cbits/win32/runProcess.c
```diff
@@ -0,0 +1,611 @@
...
+                           asynchronous))
+      goto cleanup_err;
+
+    if (!setStdHandleInfo (&sInfo.hStdOutput, _stdout, &hStdOutputRead,
+                           &hStdOutputWrite, defaultStdOutput, FALSE, TRUE,
+                           asynchronous))
+      goto cleanup_err;
+
+    if (!setStdHandleInfo (&sInfo.hStdError, _stderr, &hStdErrorRead,
+                           &hStdErrorWrite, defaultStdError, TRUE, TRUE,
```
bgamari, line 301: Shouldn't the inheritability here be similar to that of stdout?

cbits/win32/runProcess.c
```diff
@@ -0,0 +1,611 @@
...
+        && sInfo.hStdOutput !=  defaultStdOutput
+        && sInfo.hStdError  !=  defaultStdError
+        && (flags & RUN_PROCESS_IN_NEW_GROUP) == 0)
+            dwFlags |= CREATE_NO_WINDOW;   // Run without console window only when both output and error are redirected
+
+    // See #3231
+    if (close_fds
+        && _stdin == defaultStdIn
+        && _stdout == defaultStdOutput
+        && _stderr == defaultStdError) {
```
bgamari, line 315: Should it not be an error if the user sets `close_fds` yet then asks us to redirect a standard handle? I'm worried that the user might be thinking that `close_fds` gives them security which they aren't in fact getting.

System/Process/Common.hs
```diff
@@ -26,6 +26,13 @@ module System.Process.Common
 #else
     , CGid
 #endif
+
+-- WINIO is only available on GHC 8.12 and up.
+#if defined(__IO_MANAGER_WINIO__)
```
bgamari, line 31: Who is responsible for defining this macro?

cbits/win32/runProcess.c
Mistuke: I agree that this carve-out is weird, though in this case I only refactored the code and made it easier to read but left the same logic in place.  In particular this code seems to have been added as a workaround for https://gitlab.haskell.org/ghc/ghc/issues/3231 which is the thing we were still struggling with until recently. 

 The core premise being they have to wait for all the resources to be released by the process before continuing, however the issue was they waited for the main process to terminate and not all process that have inherited the handle.  This caused the same issues we were seeing with the race conditions in process spawning (just 11 years ago). 

The fix that was implemented was to not make handles inheritable if they aren't redirected. which means when the main process exits the handles are done.  This sorta works but only when you don't redirect the handles, either yourself or from the console.  It doesn't take into account when your handles are actually files and not pipes.  Which is why the bug was re-opened years ago.

This brings back to your point that maybe process jobs should be the default, after this is all said and done, we should evaluate that and make sure our current fix covers this case and we can remove this carve out.

I have created https://github.com/haskell/process/issues/186 to re-evaluate this fix, but for now I think it's easier if this PR doesn't change functionality.

Mistuke:
This should be the final commit. Everything passes with it.

snoyberg:
I'm definitely out of my depth on this one. I'm OK moving ahead with merging this if both @Mistuke and @bgamari think this is ready.

Mistuke:
Thanks  @snoyberg, I don't have anything else outstanding on this. @bgamari ?

cbits/win32/runProcess.c
bgamari: Alright, fair enough. This seems reasonable. Thanks for opening the ticket.

PR #178, opened 2020-05-30
Allow base-4.15

bgamari:


bgamari:
I suspect that revising might not be a good idea in this case as WinIO sadly may result in previos versions  breaking (see #177).

snoyberg:
Does that mean that this PR is essentially incomplete, and this commit would not be safe for release to Hackage?

bgamari:
That is correct.

snoyberg:
Understood, thanks.

bgamari:
Also, just to clarify, we do hope to get WinIO this week.

Mistuke:
process's own package bounds aren't much of an issue, process will compile with or without WINIO. The difficulty here is that GHC 8.12 itself in order to use WINIO requires at least the process it ships with.

I think the only way to solve that is for libGHC itself to have a tight version bound on process versions. But process itself can be as liberal as it wants unless I'm mistaken?

PR #179, opened 2020-05-31
testsuite: Add process-fork-wait test

bgamari:
This test verifies that the `use_process_jobs` feature works as
advertised. Specifically: on Windows `waitForProcess` should not return
until all processes created by the child (including those created with
`fork`) have exited if `use_process_jobs` is enabled.

Frankly, this test should have been added long ago.


bgamari:
Sure, I would be happy to handle it myself. Thanks for the quick feedback.

PR #182, opened 2020-06-07
Give a usable buffer to _pipe on Windows

### LINKED ISSUE #181: `createPipe` doesn't work on Windows
Example program:

```haskell
{-# LANGUAGE OverloadedStrings #-}

import qualified Data.ByteString as BS
import System.IO
import System.Process

main :: IO ()
main = do
  (h_read, h_write) <- createPipe
  let msg = "233"
  BS.hPut h_write msg
  hFlush h_write
  BS.hGet h_read (BS.length msg) >>= print
```

This prints `"233"` on Linux/macOS, but the program hangs indefinitely on Windows. Tested against `ghc-8.10.1` which ships `process-1.6.8.2`, both on my Windows 10 laptop and on [GitHub Actions](https://github.com/tweag/inline-js/runs/741183886?check_suite_focus=true).

Mistuke:
Fixes #181, the documentation seems to have been misread, the argument to `_pipe` is the number of bytes to reserve for pipe communications, not the number of pipes to create.  This ended up making pipes that can only send 2 bytes at a time.

The pipes are created in blocking mode, so any write of larger than 2 bytes block and no read returns more than 2 bytes..

Instead use 8k which is the size of the internal Buffers in `Base` which back `I/O` calls.

### DIFF
--- a/System/Process/Windows.hsc
+++ b/System/Process/Windows.hsc
@@ -308,7 +308,7 @@ createPipeInternal = do
 createPipeInternalFd :: IO (FD, FD)
 createPipeInternalFd = do
     allocaArray 2 $ \ pfds -> do
-        throwErrnoIfMinus1_ "_pipe" $ c__pipe pfds 2 (#const _O_BINARY)
+        throwErrnoIfMinus1_ "_pipe" $ c__pipe pfds 8192 (#const _O_BINARY)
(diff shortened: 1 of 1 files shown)


snoyberg:
@terrorjack Could you review if this addresses your issue?

Issue #192, opened 2020-09-07
Confusing documentation for createProcess_


tomjaguarpaw:
I'm really confused by the documentation for `createProcess_`.  For reference, the documentation for `createProcess` (without underbar) says

```
createProcess
    :: CreateProcess
    -> IO (Maybe Handle, Maybe Handle, Maybe Handle, ProcessHandle) 
```

"... createProcess returns (mb_stdin_hdl, mb_stdout_hdl, mb_stderr_hdl, ph)"

and the documentation for `createProcess_` (with underbar) says

```
createProcess_
    :: String
    -> CreateProcess
    -> IO (Maybe Handle, Maybe Handle, Maybe Handle, ProcessHandle)
```

"This function is almost identical to `createProcess`. The only differences are:

*    Handles provided via UseHandle are not closed automatically.
*    This function takes an extra String argument to be used in creating error messages.
*    use_process_jobs can be set in CreateProcess since 1.5.0.0 in order to create an I/O completion port to monitor a process tree's progress on Windows.

The function also returns two new handles: * an I/O Completion Port handle on which events will be signaled. * a Job handle which can be used to kill all running processes."

But where are these two new handles?  Are they two of the three returned in the tuple from the function?  If so then those three are not stdin, stdout, and stderr like they are for `createProcess`, so `createProcess_` is *not* "almost identical to `createProcess`". If they are not part of that tuple then where are they? Something needs to be clarified here.

snoyberg:
This looks to me like an accidental merging of two functions' docs, see:

https://github.com/haskell/process/commit/3a5935ce6353e5c9e35214a070c31342098e5c6c#diff-c5db4cc90b03efcbad68033c8d800d7bL73

@Mistuke would you mind having a look and seeing if that's an accurate description?

Mistuke:
Hmm it looks like I wrote the docs without taking into account the surrounding context.  The doc is accurate, but the last two points aren't a difference with `createProcess` as the same behavior can be gotten with `createProcess` by setting `use_process_jobs`.

> * use_process_jobs can be set in CreateProcess since 1.5.0.0 in order to create an I/O completion port to monitor a process tree's progress on Windows. 

This and the part below it should also be under the docs for `createProcess`.

> But where are these two new handles? Are they two of the three returned in the tuple from the function?

They are in `ProcessHandle`. when `use_process_jobs` is used you get a `OpenExtHandle` back instead of `OpenHandle`. The former gives you two handles instead of one.

[...]

PR #193, opened 2020-09-07
Fix Haddock

tomjaguarpaw:
Fixes https://github.com/haskell/process/issues/192

@Mistuke: could you please check whether I have interpreted [your comment](https://github.com/haskell/process/issues/192#issuecomment-688490600) correctly?


tomjaguarpaw:
Hmm, I think the following also belongs on `createProcess`, doesn't it?  Let me know and I can add it to this PR.

```
-- * 'use_process_jobs' can be set in CreateProcess since 1.5.0.0 in order to create
--   an I/O completion port to monitor a process tree's progress on Windows.
```

System/Process.hs
```diff
@@ -200,6 +200,15 @@ always the desired behavior. In cases where you would like to leave the
...
 to be in binary mode then use 'hSetBinaryMode'.
 
+'use_process_jobs' can be set in CreateProcess since 1.5.0.0 in order to create
+an I/O completion port to monitor a process tree's progress on Windows.
+The function returns two new handles (inside @/ph/@):
+
+  * an I/O Completion Port handle on which events will be signaled.
+  * a Job handle which can be used to kill all running processes.
+
+On POSIX platforms these two new handles will always be Nothing
```
Mistuke, line 210: On posix you never see the `OpenExt` format, so you always only have the single handle. (you still get the process handle).

So to rephrase on
Posix (regardless of what `use_process_jobs` is) you get one handle (the process handle).
On windows depending on the value of `use_process_jobs` you get one or two handles.

System/Process.hs
```diff
@@ -200,6 +200,15 @@ always the desired behavior. In cases where you would like to leave the
 instead. All created @Handle@s are initially in text mode; if you need them
 to be in binary mode then use 'hSetBinaryMode'.
 
+'use_process_jobs' can be set in CreateProcess since 1.5.0.0 in order to create
+an I/O completion port to monitor a process tree's progress on Windows.
+The function returns two new handles (inside @/ph/@):
+
+  * an I/O Completion Port handle on which events will be signaled.
```
Mistuke, line 207: This is no longer true. We don't probe for events anymore so this handle is no longer given. Comment is outdated.

Mistuke (commented):
Thanks @tomjaguarpaw just some minor changes.

System/Process.hs
tomjaguarpaw: Is the following wrong then? It is still talking about events.

https://hackage.haskell.org/package/process-1.6.10.0/docs/System-Process-Internals.html#t:ProcessHandle__

> On Windows a second wait method can be used to block for event completion. This requires two handles. A process job handle and a events handle to monitor.

Also, should this comment be attached to `ProcessHandle` rather than `ProcessHandle__`?  The latter is internal.

tomjaguarpaw:
OK, changes made.

System/Process.hs
tomjaguarpaw: I did the `ProcessHandle`/`ProcessHandle__` Haddock switch at https://github.com/haskell/process/pull/194

System/Process.hs
```diff
@@ -200,6 +200,13 @@ always the desired behavior. In cases where you would like to leave the
 instead. All created @Handle@s are initially in text mode; if you need them
 to be in binary mode then use 'hSetBinaryMode'.
 
+@/ph/@ contains a handle to the running process.  On Windows
+'use_process_jobs' can be set in CreateProcess in order to create an
```
Mistuke, line 204: you removed the wrong one :) This should be

```
@/ph/@ contains a handle to the running process.  On Windows
'use_process_jobs' can be set in CreateProcess in order to create an
I/O completion port to monitor a process tree's progress.  If it is
set then that port is also returned inside @/ph/@.  The port can be
used to kill all running subprocesses.  This feature has been
available since 1.5.0.0.
```

We simply no longer use completion ports to monitor for processes.
We only use a job object. See `waitForJobCompletion`. So this should be
something like

```
@/ph/@ contains a handle to the running process.  On Windows
'use_process_jobs' can be set in CreateProcess in order to create a
Win32 Job object to monitor a process tree's progress.  If it is
set then that job is also returned inside @/ph/@.  The job and process
can be used to kill all running sub-processes.  This feature
has been available since 1.5.0.0.
```

Though realistically speaking this is an implementation detail. The user is expected to
only use `ph`, e.g. `terminateProcess` takes `ph` not the job.

I'm not sure how much of this we should mention or should stay internal @snoyberg ?

tomjaguarpaw: Fixed accordingly.

I sympathise with the wish to not say too much about implementation details but as things stand the documentation is simply wrong. Could we at least merge this PR and then think about how much of the internals are mentioned in a separate PR?

Mistuke:
Thanks for your patience. This looks OK to me (you have a small typo, missing space between feature and has) but you need @snoyberg to merge.

tomjaguarpaw:
> you have a small typo, missing space between feature and has

Ah yes, thanks, fixed.

Mistuke (approved):

snoyberg (approved):
Thanks!

PR #194, opened 2020-09-09
Move Haddock from ProcessHandle__ to ProcessHandle

tomjaguarpaw:
I think this Haddock should be moved to the externally supported data definition.  Is that right?

### DIFF
--- a/System/Process/Common.hs
+++ b/System/Process/Common.hs
@@ -189,6 +189,15 @@ data StdStream
 -- ----------------------------------------------------------------------------
 -- ProcessHandle type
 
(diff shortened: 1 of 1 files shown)


tomjaguarpaw:
(The diff looks like the definition has moved, but it's the same effect as moving the Haddock)

tomjaguarpaw:
Pinging @snoyberg: I think this is a fairly straightforward change.

snoyberg:
Sorry for the delay, I completely missed this. Thanks for the ping and the change.

tomjaguarpaw:
You're welcome. Thanks for your maintainership of this great library!

PR #199, opened 2020-12-18
GH196: Fix terminateJob

### LINKED ISSUE #196: use_process_jobs doesn't work with terminateProcess
```haskell
import System.Process

main :: IO ()
main = do
  (_, _, _, ph) <-
    createProcess $
      (proc "sleep" ["21600"]) {use_process_jobs = True}
  terminateProcess ph
  print =<< waitForProcess ph
```

On Windows, this may either hang indefinitely or crash with `thread blocked indefinitely in an MVar operation`. See https://github.com/TerrorJack/repro-201111/runs/1385388481?check_suite_focus=true for the CI log of the repro, using Windows 2019/ghc-8.8.4/process-1.6.9.0.

Mistuke:
Hi,

`terminateJob` is being called from within `terminateProcess` which already has a lock on the `MVar`. 
This add a new function `terminateJobUnsafe` which does not take an `MVar` and so is not thread-safe but can be called from `terminateProcess` safely.

Fixes #196

### DIFF
--- a/System/Process.hs
+++ b/System/Process.hs
@@ -817,7 +817,7 @@ terminateProcess ph = do
     case p_ of
       ClosedHandle  _ -> return ()
 #if defined(WINDOWS)
-      OpenExtHandle{} -> terminateJob ph 1 >> return ()
+      OpenExtHandle{} -> terminateJobUnsafe p_ 1 >> return ()
 #else
       OpenExtHandle{} -> error "terminateProcess with OpenExtHandle should not happen on POSIX."
 #endif
--- a/System/Process/Internals.hs
+++ b/System/Process/Internals.hs
@@ -41,6 +41,7 @@ module System.Process.Internals (
     unwrapHandles,
 #ifdef WINDOWS
     terminateJob,
+    terminateJobUnsafe,
     waitForJobCompletion,
     timeout_Infinite,
 #else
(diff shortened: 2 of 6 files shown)


snoyberg:
The AppVeyor tests failed on this build.

Side note: I think I'll move CI over to Github Actions instead of the Travis/AppVeyor split, just an FYI.

Mistuke:
> The AppVeyor tests failed on this build.
> 

That test is somewhat broken which is why it sometimes passes and sometimes fails. See https://sourceforge.net/p/mingw/mailman/mingw-users/thread/CAO2ddnbqqTwABiLSpyDN-ynyQR%3D6fRt5Mt94gszkgQoJF-atOA%40mail.gmail.com/#msg30566147

In general there's no guarantee that the PID inside a Cygwin process is the same as the PID in Windows. Cygwin is trying to emulate a posix environment in user space.  It will try to have the PIDs match but this may not always succeed which is why this test is flaky.

In any case, I've updated the test to read the `winpid` from `procfs` instead.

Mistuke:
@snoyberg alright, the test is fixed to work properly on Windows now.

snoyberg (changes requested):
Looks great. One more request: could you add a changelog entry explaining the change?

Mistuke:
Done.

snoyberg (approved):

snoyberg:
Thanks!

PR #202, opened 2021-02-03
Bump Win32 upper bound to accomodate Win32-2.10

bgamari:
This should be the last change necessary for GHC 9.0. No new release is necessary but a Hackage revision would likely be best.

Thanks again!


snoyberg:
Revised on Hackage

PR #204, opened 2021-03-21
WIP: make waitForProcess async safe

snoyberg:
Moving from https://github.com/fpco/typed-process/issues/38. CC @norfairking

This PR is not complete, the tests fail. I'm trying to determine whether this is a reasonable approach. My understanding of the issue here:

1. `waitForProcess` wants to allow the C FFI call `c_waitForProcess` to be interruptible, so that `waitpid` can be interrupted
2. If an async exception lands at exactly the wrong time, the `waitpid` call with succeed, but the FFI call will throw an exception due to the async exception
3. In this case, the system process table no longer has an entry for the given PID, but the `MVar` inside `ProcessHandle` believes that the handle is still open, allowing future `waitForProcess` calls to occur
4. Future calls throw an exception, since it's using an unknown PID

My best guess is that, in order to both allow `waitpid` to be interrupted _and_ to reliably capture the exit code if it's generated, we need to:

1. No longer rely on return codes from `c_waitForProcess`, since those will be lost in the case of an async exception
2. Therefore, we need to detect whether the `waitpid` call succeeded or not by passing in another `int*` (along with the existing output parameter `pret`).
3. Even if `c_waitForProcess` throws an exception, we need to update the `MVar` with the exit code, if it's returned successfully

I've taken a stab at this in this PR. @simonmar I was hoping you could have a look and let me know if this approach works, and/or if there's a better way to attack this kind of problem.

### DIFF
--- a/System/Process.hs
+++ b/System/Process.hs
@@ -680,9 +680,14 @@ waitForProcess ph@(ProcessHandle _ delegating_ctlc _) = lockWaitpid $ do
   p_ <- modifyProcessHandle ph $ \p_ -> return (p_,p_)
   case p_ of
     ClosedHandle e -> return e
-    OpenHandle h  -> do
+    -- Begin by masking all async exceptions, and only allowing them
+    -- during the actual C FFI call below.
+    OpenHandle h  -> mask $ \restore -> do
         -- don't hold the MVar while we call c_waitForProcess...
-        e <- waitForProcess' h
+        (e, eres) <- waitForProcess' h restore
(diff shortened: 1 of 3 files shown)


NorfairKing:
@snoyberg Thanks for CC-ing me. I don't feel qualified to review this PR, to be honest. I'm betting that @nh2 will, if he has some spare cycles to help with this.

nh2:
I thought a bit about it and will try to reply to reply tomorrow.

NorfairKing:
@simonmar Gentle ping about this.

PR #205, opened 2021-06-27
Implement getCurrentPid

yutotakano:
### Change

This PR adds a new exported function in System.Process: `getCurrentPid`.

This calls `getCurrentProcessId` from System.Win32.Process on Windows, and `getProcessID` from System.Posix.Process on POSIX systems.

### Rationale

This seems like an oversight considering `getPid` is implemented already. I sometimes require the current running program's process ID. Currently, I have to do the CPP conditional logic myself every time, which is a pain. Putting this in `process` seems natural. 

Furthermore, my understanding of `process` is that it is a unified interface to System.Win32.Process and System.Posix.Process. As both modules provide a function to "get the currently executing process id", I feel it makes sense to have a cross-platform wrapper function.

### Remaining Issues

The `@since` field has been left as a TODO since I'm not sure what version this will be merged in. 

### Tests

I could not come up with a method to test the PR. Unlike with the test for `getPid`, there's nothing to compare the value against, I think. It's worked from my past experiences in personal projects.

I wasn't able to ensure a stable setup to test to begin with, as running the existing test suite `stack test` on the Windows version of Stack failed due to several reasons (`getPid` failing, and `detatch_console` failing). Any feedback or insight is therefore appreciated.

### DIFF
--- a/System/Process.hs
+++ b/System/Process.hs
@@ -49,6 +49,7 @@ module System.Process (
     showCommandForUser,
     Pid,
     getPid,
+    getCurrentPid,
 
     -- ** Control-C handling on Unix
     -- $ctlc-handling
@@ -93,8 +94,9 @@ import System.IO
 import System.IO.Error (mkIOError, ioeSetErrorString)
 
 #if defined(WINDOWS)
-import System.Win32.Process (getProcessId, ProcessId)
+import System.Win32.Process (getProcessId, getCurrentProcessId, ProcessId)
 #else
+import System.Posix.Process (getProcessID)
 import System.Posix.Types (CPid (..))
 #endif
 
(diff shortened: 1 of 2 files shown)


snoyberg:
Thanks for the PR and the catch about the unstable `stack test`. I've pushed some commits to move over to GitHub Actions and temporarily disable the broken tests. Can you rebase against master?

yutotakano:
I've confirmed the remaining tests now pass perfectly on both my Windows and Linux Haskell setups. I've also rebased and force-pushed the changes. Thanks for the quick reply!

System/Process.hs
```diff
@@ -652,6 +654,24 @@ getPid (ProcessHandle mh _ _) = do
...
 
+-- ----------------------------------------------------------------------------
+-- getCurrentPid
+
+-- | Returns the PID (process ID) of the current process. On POSIX systems,
+-- this calls 'getProcessID' from "System.Posix.Process" in the @unix@ package.
+-- On Windows, this calls 'getCurrentProcessId' from "System.Win32.Process" in
+-- the @Win32@ package.
+--
+-- @since TODO
```
snoyberg, line 665: We'll call this 1.6.12.0. Can you update this, as well as add an entry to the changelog? Other than that it looks ready to go, thank you!

yutotakano: Done!

yutotakano:
Huh, didn't mean to tag the user, sorry SInCE!

snoyberg (approved):
Thank you!

PR #217, opened 2021-10-13
Fix #216: more documentation of `String` argument of `createProcess_`

### LINKED ISSUE #216: `String` argument to `createProcess_` insufficiently documented
https://github.com/haskell/process/blob/b39bbc0625c99c8c02840d8fd3ae45f062c9c78a/System/Process/Internals.hs#L79-L80

This does not tell me what to pass as this `String` when calling `createProcess_`:
* When will it be used exactly (for which errors) and how will it be rendered in the error message?
* Examples would help.

If I browse the source code, the caller name is passed in here, but this seems to only make sense for in-library uses.  
E.g. what would it help a user if in the error message they are presented with the name of a function in my Haskell sources?

andreasabel:
Fix #216: more documentation of `String` argument of `createProcess_`

### DIFF
--- a/System/Process/Internals.hs
+++ b/System/Process/Internals.hs
@@ -88,7 +88,7 @@ createProcess_
   :: String
        -- ^ Function name (for error messages).
        --
-       --   This is can be any 'String', but will typically be the name of the caller.
+       --   This can be any 'String', but will typically be the name of the caller.
        --   E.g., 'spawnProcess' passes @"spawnProcess"@ here when calling
        --   'createProcess_'.
   -> CreateProcess


System/Process/Internals.hs
```diff
@@ -85,7 +85,12 @@ import System.Process.Posix
 --
 -- @since 1.2.1.0
 createProcess_
-  :: String                     -- ^ function name (for error messages)
+  :: String
+       -- ^ Function name (for error messages).
+       --
+       --   This is can be any 'String', but will typically be the name of the caller.
```
snoyberg, line 91:
```suggestion
       --   This can be any 'String', but will typically be the name of the caller.
```

PR #222, opened 2021-12-17
Remove reqlib modifier from ghc test

mpickering:
The reqlib modifier is going to be removed from the ghc testsuite.


snoyberg:
Thanks! Feel free to push test suite changes directly to master if that’s easier for you in the future.

PR #228, opened 2022-01-11
posix: Ensure that errno is set after posix_spawnp fails

### LINKED ISSUE #227: Process 1.6.13 changed observable behaviour
Reproducer is https://github.com/angerman/repro-process-behaviour-change-1.6.13

Will produce the following output:
```sh
$ make all CABAL=$(which cabal) GHC=$(which ghc)
Building against process < 1.6.13 ...
Running against process < 1.6.13; with an empty PATH, this should result in a git not found message
<no git executable found>
Building against process >= 1.6.13 ...
Running against process >= 1.6.13; this now throws an exception instead.
process-crash: git: readCreateProcessWithExitCode: posix_spawnp: failed (Undefined error: 0)
make: *** [Makefile:9: all] Error 1
```

We are now throwing an exception where we didn't before.

bgamari:
In contrast to glibc, Darwin's posix_spawnp implementation doesn't set
errno on failure.

Fixes #227.

### DIFF
--- a/cbits/posix/posix_spawn.c
+++ b/cbits/posix/posix_spawn.c
@@ -201,6 +201,7 @@ do_spawn_posix (char *const args[],
 
     r = posix_spawnp(&pid, args[0], &fa, &sa, args, environment ? environment : environ);
     if (r != 0) {
+        errno = r; // posix_spawn doesn't necessarily set errno; see #227.
         *failed_doing = "posix_spawnp";
         goto fail;
     } else {


bgamari:
@angerman says that this fixes the issue for them.

angerman:
I would say this might even fix all the other libc implementations that do not implment glibc's behaviour. E.g. BSD, Android, musl, ...

tmcgilchrist:
This should fix FreeBSD as well based on the manpage (I don't have a FreeBSD machine to test it on)

```
RETURN VALUES
     Upon successful completion, posix_spawn() and posix_spawnp() return the
     process ID	of the child process to	the parent process, in the variable
     pointed to	by a non-NULL pid argument, and	return zero as the function
     return value.  Otherwise, no child	process	is created, no value is	stored
     into the variable pointed to by pid, and an error number is returned as
     the function return value to indicate the error.  If the pid argument is
     a null pointer, the process ID of the child is not	returned to the
     caller.
```
from https://www.freebsd.org/cgi/man.cgi?query=posix_spawn&sektion=3

OpenBSD sets an errno so the old behaviour was fine.

```
In case of an error, both functions may return fork() or exec() return values and set errno accordingly.
```
from https://man.openbsd.org/posix_spawn.3

snoyberg:
Thanks! I've merged and added a changelog entry.

PR #229, opened 2022-01-11
Fix #219

### LINKED ISSUE #219: Bad address when using relative cwd and setting env
I started getting a `Bad address` error when executing processes after updating from `lts-18.5` to `lts-18.10`, which brings `process` from `v1.6.9.0` to `v1.6.13.2`. It seems to be related to setting `cwd` and `env` in the `CreateProcess`.

Everything works fine if you don't change `env`. You can touch `env`, if you don't change `cwd` too. You can change `env` and change `cwd`, but depending on what you are changing `cwd` to and how that affects the command. For example, `cwd = /bin` and a command of `./true` works fine even when changing `env`. Unfortunately, I have not been able to narrow down the pattern in that last variability. Note that setting `cwd` to `.` (or the full path to `.`) does not count as "change" for this issue, but setting `env` to the result of `getEnvironment` does. Straight-forward, right?

So, there are a lot of different combinations that trigger it, but here is one minimal repro:

```console
% tree
.
├── Main.hs
└── sub
    └── exe

1 directory, 2 files
```

```console
% cat sub/exe
#!/bin/sh    
true
```

```hs
-- Main.hs
{-# LANGUAGE TypeApplications #-}

module Main where

import Prelude

import Control.Exception (SomeException, try)
import System.Environment (getEnvironment)
import System.Process

main :: IO ()
main = do
    currentEnv <- getEnvironment
    let cp = (proc "./exe" []) { cwd = Just "./sub", env = Just currentEnv }
    x <- try @SomeException $ withCreateProcess cp $ \_ _ _ -> waitForProcess
    print x
```

Works in 18.5, not in 18.10:

```console
% stack --resolver lts-18.5 runhaskell Main.hs
Right ExitSuccess
% stack --resolver lts-18.10 runhaskell Main.hs
Left ./exe: createProcess: exec: failed (Bad address)
```

Note that it is indeed isolated to `process`: I can use `18.10` with an `extra-dep` of `process-1.6.9.0` and the bug goes away. I'm just not showing it that way because it requires more setup to repro.

I experienced this in `typed-process` (using `setWorkingDir` and `setEnv`), but eventually found the bug to be here.

bgamari:
This fixes #219 by ensuring that `find_executable` respects the working directory requested by the user.

I also fix an orthogonal bug which meant that `execvpe` wouldn't be used, even if available.

### DIFF
--- a/cbits/posix/find_executable.c
+++ b/cbits/posix/find_executable.c
@@ -82,7 +82,9 @@ char *find_executable(char *filename) {
     }
 
     char *search_path = get_executable_search_path();
-    return find_in_search_path(search_path, filename);
+    char *result = find_in_search_path(search_path, filename);
+    free(search_path);
+    return result;
 }
 
 #endif
--- a/cbits/posix/find_executable.c
+++ b/cbits/posix/find_executable.c
@@ -14,16 +14,46 @@
 // the below is only necessary when we need to emulate execvpe.
 #if !defined(HAVE_execvpe)
 
-/* Return true if the given file exists and is an executable. */
-static bool is_executable(const char *path) {
-    return access(path, X_OK) == 0;
+/* A quick check for whether the given path is absolute. */
+static bool is_absolute(const char *path) {
+    return path[0] == '/';
+}
+
+static char *concat_paths(const char *path1, const char *path2) {
+    if (is_absolute(path2)) {
+        return strdup(path2);
+    } else {
+        int len = strlen(path1) + 1 + strlen(path2) + 1;
+        char *tmp = malloc(len);
+        int ret = snprintf(tmp, len, "%s/%s", path1, path2);
+        if (ret < 0) {
+            free(tmp);
+            return NULL;
+        }
+        return tmp;
+    }
+}
+
+/* Return true if the given file exists and is an executable, optionally
+ * relative to the given working directory.
+ */
+static bool is_executable(char *working_dir, const char *path) {
+    if (working_dir && !is_absolute(path)) {
+        char *tmp = concat_paths(working_dir, path);
+        bool ret = access(tmp, X_OK) == 0;
+        free(tmp);
+        return ret;
+    } else {
+        return access(path, X_OK) == 0;
+    }
 }
 
 /* Find an executable with the given filename in the given search path. The
  * result must be freed by the caller. Returns NULL if a matching file is not
  * found.
   */
-static char *find_in_search_path(char *search_path, const char *filename) {
+static char *find_in_search_path(char *working_dir, char *search_path, const char *filename) {
+    int workdir_len = strlen(working_dir);
     const int filename_len = strlen(filename);
     char *tokbuf;
     char *path = strtok_r(search_path, ":", &tokbuf);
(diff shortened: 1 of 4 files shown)


cbits/posix/find_executable.c
```diff
@@ -12,18 +12,48 @@
...
+    return path[0] == '/';
+}
+
+static char *concat_paths(const char *path1, const char *path2) {
+    if (is_absolute(path2)) {
+        return strdup(path2);
+    } else {
+        int len = strlen(path1) + 1 + strlen(path2) + 1;
+        char *tmp = malloc(len);
+        int ret = snprintf(tmp, len, "%s/%s", path1, path2);
```
snoyberg, line 28: This will end up with a double slash in the case that `path1` ends in a slash. I don't think this has any negative impact on the behavior, however.

cbits/posix/find_executable.c
```diff
@@ -12,18 +12,48 @@
...
+        }
+        return tmp;
+    }
+}
+
+/* Return true if the given file exists and is an executable, optionally
+ * relative to the given working directory.
+ */
+static bool is_executable(char *working_dir, const char *path) {
+    if (working_dir) {
```
snoyberg, line 41: Slight potential performance improvement: check if `path` is absolute here to avoid the unnecessary `strdup`.

cbits/posix/find_executable.c
```diff
@@ -34,13 +64,13 @@ static char *find_in_search_path(char *search_path, const char *filename) {
...
 #endif
-        const int tmp_len = filename_len + 1 + strlen(path) + 1;
+        const int tmp_len = workdir_len + 1 + filename_len + 1 + strlen(path) + 1;
 #if defined(__GNUC__) && __GNUC__ == 6 && __GNUC_MINOR__ == 3
 #pragma GCC diagnostic pop
 #endif
         char *tmp = malloc(tmp_len);
-        snprintf(tmp, tmp_len, "%s/%s", path, filename);
-        if (is_executable(tmp)) {
+        snprintf(tmp, tmp_len, "%s/%s/%s", working_dir, path, filename);
```
snoyberg, line 72: Does this behave correctly in the case that `path` is an absolute path?

cbits/posix/find_executable.c
```diff
@@ -74,15 +104,36 @@ static char *get_executable_search_path(void) {
...
-    if (strchr(filename, '/') && is_executable(filename)) {
+/* Find the given executable in the executable search path relative to
+ * workingDirectory (or the current directory, if NULL).
+ * N.B. the caller is responsible for free()ing the result.
+ */
+char *find_executable(char *working_dir, char *filename) {
+    /* Drop trailing slash from working directory if necessary */
+    if (working_dir) {
+        int workdir_len = strlen(working_dir);
+        if (working_dir[workdir_len] == '/') {
```
snoyberg, line 115: I think this needs to check `working_dir[workdir_len - 1]`, right?

cbits/posix/find_executable.c
bgamari: Indeed, double-slashes shouldn't affect the meaning of the path.

cbits/posix/find_executable.c
bgamari: Quite right!

cbits/posix/find_executable.c
bgamari: Fair point. I'll implement.

cbits/posix/find_executable.c
bgamari: No, I believe it does not. Good catch!

snoyberg (approved):

Issue #226, opened 2022-01-10
waitForProcess: does not exist (No child processes)


dio4ev:
Hi all!

I hope that this is the right place to ask such a question.

I try to parallize parts of a program of mine, but get a strange runtime error and as far as I know it is related to something in the process package.
In my tool, I get some specific input, build a constraint from it which can be solved in 3 different ways (3 solvers; involving external tools).
Thus I would like to run these three different solvers on the constraint in parallel and return the result of the fastest one. 
In simple words it looks as follows:
A solver `s` has a type `s :: constraint -> IO [a]` and I apply the constraint to all solvers to get a list of possible solutions `[IO [a]]`. After that I call `async` [0] on them and wait with `waitAny` for a result (if I find a correct result, i.e. a non-empty list is returned, then I kill all the other jobs with `cancel`). Within a solver external tools are called with `readProcess` and `readProcessWithExitCode`. **The problem is that I get quite often the following runtime error `waitForProcess: does not exist (No child processes)`.**

* Is mixing `async` and `process` causing this?

* Is this related to something similar as [1]?

* Does anybody has an idea what could cause this?

In the meantime I replaced the `async` package functions completely by `forkIO` etc., but still this exception occurs.
However, I never experienced this exception before I parallelized my program.

Thank you very much and cheers! 

[0]: https://hackage.haskell.org/package/async-2.2.4
[1]: https://github.com/haskell/process/issues/107

Fuuzetsu:
It's a bit hard to help without seeing any code. You should be able to mix `async` and `process`, there's no fundamental limitation I think. #107 is probably not related.

I assume the code is private but if you could strip out the process invocation/waiting/cancel parts and post them, it'd be a lot easier to give help. Preferably the non-`async` version, just fewer moving blocks.

dio4ev:
Sure I understand that, I will try to strip out some of the important parts:

After the input is parsed I call all elements of `solvers` on the input with the following parallelization:
```
ecParP :: Input -> Maybe Int -> Maybe Int -> IO [Result]
ecParP p i n = do
  var <- newEmptyMVar
  threads <- mapM (forkIO . mVarFun null var . run') solvers
  waitThreadSet [] success (S.fromList threads) var
  where (_, c) = getMain p
        run' solv = run solv c

        success Nothing = False
        success (Just ress) = any succ' ress
          where
            succ' (Failure _ _) = False
            succ' (Success _ _) = True
```
```
[...]

PR #231, opened 2022-02-03
Fix waitForProcess not closing process handles with delegate_ctlc

robx:
I believe this might address #226, which I've run across myself when interrupting processes spawned with `delegate_ctlc` within `withCreateProcess`. It turns out this isn't a race condition (though I'm not saying there are none), it's just that the first `withCreateProcess` call fails to close the handle when it throws `UserInterrupt`.

There's a test case that covers the original behaviour I saw (an uncaught exception when the subprocess is interrupted), and then a couple of test cases narrowing down to the actual bug.

I haven't tested this on Windows, and I'm not sure to what extent the tests make sense there -- we might need to adapt or skip them.

### DIFF
--- a/System/Process.hs
+++ b/System/Process.hs
@@ -730,6 +730,7 @@ waitForProcess ph@(ProcessHandle _ delegating_ctlc _) = lockWaitpid $ do
               closePHANDLE ph'
               closePHANDLE job'
               return (ClosedHandle e, e)
+        -- omit endDelegateControlC since it's a no-op on Windows
         return e'
 #else
     OpenExtHandle _ _job ->
--- a/System/Process.hs
+++ b/System/Process.hs
@@ -703,15 +703,15 @@ waitForProcess ph@(ProcessHandle _ delegating_ctlc _) = lockWaitpid $ do
     OpenHandle h  -> do
         -- don't hold the MVar while we call c_waitForProcess...
         e <- waitForProcess' h
-        e' <- modifyProcessHandle ph $ \p_' ->
+        (e', was_open) <- modifyProcessHandle ph $ \p_' ->
           case p_' of
-            ClosedHandle e' -> return (p_', e')
+            ClosedHandle e' -> return (p_', (e', False))
             OpenExtHandle{} -> fail "waitForProcess(OpenExtHandle): this cannot happen"
             OpenHandle ph'  -> do
               closePHANDLE ph'
-              when delegating_ctlc $
-                endDelegateControlC e
-              return (ClosedHandle e, e)
+              return (ClosedHandle e, (e, True))
+        when (was_open && delegating_ctlc) $
+          endDelegateControlC e
         return e'
 #if defined(WINDOWS)
     OpenExtHandle h job -> do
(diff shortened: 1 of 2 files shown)


test/main.hs
```diff
@@ -86,54 +109,148 @@ main = do
...
+                Nothing -> error $ "waitForProcess threw: " <> show e
+            Right _ -> error $ "expected exception, got " <> show res
+        _ -> case res of
+            Left e -> error $ "waitForProcess threw: " <> show e
+            Right ExitSuccess -> error "waitForProcess: expected failure code"
+            _ -> putStrLn "result ok"
+
+    checkSecond :: String -> Bool -> Either SomeException ExitCode -> IO ()
+    checkSecond sig delegate res = case (sig, delegate) of
+        ("INT", True) -> checkFirst "INT" False res -- TODO: should this also throw?
```
robx, line 251: I *think* this is fine, since multiple `waitForProcess` calls would be happening in contexts where we're bailing out anyway.

System/Process.hs
```diff
@@ -703,15 +703,15 @@ waitForProcess ph@(ProcessHandle _ delegating_ctlc _) = lockWaitpid $ do
     OpenHandle h  -> do
         -- don't hold the MVar while we call c_waitForProcess...
         e <- waitForProcess' h
-        e' <- modifyProcessHandle ph $ \p_' ->
+        (e', was_open) <- modifyProcessHandle ph $ \p_' ->
```
snoyberg, line 706: I'm honestly not following what the difference is in the previous and new version of the code, can you elaborate? It seems like the goal is to call `endDelegateControlC` outside of `modifyProcessHandle`, but I'm not sure why that makes a difference.

robx: The thing is that `endDelegateControlC` throws `UserInterrupt`, which means we're not setting the process handle to "closed". (`modifyProcessHandle` is just `modifyMvar`, which leaves the value as is if an exception is raised.)

Essentially, before was:

```
start modifyProcessHandle
  endDelegateControlC
    throw UserInterrupt
end modifyProcessHandle, actually writing handle (never reached!)
```

After is:

```
start modifyProcessHandle
end modifyProcessHandle, actually writing handle
endDelegateControlC
  throw UserInterrupt
```

snoyberg: Got it. Could you add a comment to the `endDelegateControlC` call about the fact that it throws `UserInterupt`. I'd actually not noticed that personally (thus my question), and it would be a good thing to help avoid someone accidentally "improving" the code in the future.

robx:
I pushed some fixes for the CI failures. Once they pass I'd still like to clean up the commits a little.

System/Process.hs
```diff
@@ -725,8 +729,6 @@ waitForProcess ph@(ProcessHandle _ delegating_ctlc _) = lockWaitpid $ do
             OpenExtHandle ph' job' -> do
               closePHANDLE ph'
               closePHANDLE job'
-              when delegating_ctlc $
```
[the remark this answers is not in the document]
robx: Yeah, see also the commit message:

> Additionally, remove endDelegateControlC entirely from the Windows-only
> OpenExtHandle branch, where it's a no-op and was also in the wrong place.

It seemed wrong to leave the misleading previous version, but also weird to write some
no-op version of the fix here since `delegating_ctlc` has no effect on Windows.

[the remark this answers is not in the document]
snoyberg: I think an explicit comment in the code explaining why it's not needed would be a good thing. Otherwise someone may think it was just an oversight and add it back in the future.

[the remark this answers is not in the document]
robx: this is done now

snoyberg (approved):
Awesome, thank you!

PR #234, opened 2022-02-20
Phyx/winio permissions and mode fixes

Mistuke:
Hi,

This PR fixes several issues:

1. The permissions for the pipes created through `createPipe` were accidentally swapped. The `read` end got `write` permissions while the `write` part got `read` permissions in Haskell.  The permissions were correct in the underlying OS object but this caused a very weird message when the handles were used.  This corrects them so they're in sync.
2.  It documents a requirement for users of `createPipe` in that since we don't know who's going to use which part of the handles that it's the caller's responsibility to associate the handles with the correct I/O manager.   This documents an example on how to do so.
3. Pipes can be opened in two modes, `message` and `byte`, essentially `chunk` and `streaming` mode.   `CreateNamedPipeW` defaults to `message` mode while `CreateFile` defaulted to `byte` mode.  However the modes need to match.  This corrects it by always setting both modes explicitly ensuring that both ends always match.  For these pipes we default to `chunk` instead  of  `streaming` mode as things such as `iserv` require the entire message in order to continue anyway.

This fixes the last issues to enable the new I/O manager with iserv support in GHC.

/cc @bgamari

### DIFF
--- a/System/Process/Internals.hs
+++ b/System/Process/Internals.hs
@@ -187,7 +187,7 @@ runGenProcess_ fun c _ _ = createProcess_ fun c
 -- When this function is used with WinIO enabled it's the caller's
 -- responsibility to register the handles with the I/O manager.
 -- If this is not done the operation will deadlock.  Association can
--- be done as follow:
+-- be done as follows:
 --
 -- @
 --     #if defined(__IO_MANAGER_WINIO__)
--- a/System/Process/Internals.hs
+++ b/System/Process/Internals.hs
@@ -182,6 +182,33 @@ runGenProcess_ fun c _ _ = createProcess_ fun c
 -- | Create a pipe for interprocess communication and return a
 -- @(readEnd, writeEnd)@ `Handle` pair.
 --
+-- * WinIO Support
+--
+-- When this function is used with WinIO enabled it's the caller's
+-- responsibility to register the handles with the I/O manager.
+-- If this is not done the operation will deadlock.  Association can
+-- be done as follow:
+--
+-- @
+--     #if defined(__IO_MANAGER_WINIO__)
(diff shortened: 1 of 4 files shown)


System/Process/Internals.hs
```diff
@@ -182,6 +182,33 @@ runGenProcess_ fun c _ _ = createProcess_ fun c
 -- | Create a pipe for interprocess communication and return a
 -- @(readEnd, writeEnd)@ `Handle` pair.
 --
+-- * WinIO Support
+--
+-- When this function is used with WinIO enabled it's the caller's
+-- responsibility to register the handles with the I/O manager.
+-- If this is not done the operation will deadlock.  Association can
+-- be done as follow:
```
snoyberg, line 190:
```suggestion
-- be done as follows:
```

snoyberg (commented):
LGTM, though I only superficially understand this. Does this need a more thorough review from others? And does this warrant an immediate release, or is it acceptable to wait till the next release?

Mistuke:
> LGTM, though I only superficially understand this. Does this need a more thorough review from others? And does this warrant an immediate release, or is it acceptable to wait till the next release?

Cheers! Thanks for taking a look! Hmm we can wait a day or two see if @bgamari has any comments, but other than that it fixes the last bits of errors in the GHC testsuite when the new I/O manager is enabled by default.

I don't think it needs an immediate release, we're aiming to enable the I/O manager by default in GHC 9.4, so as long as the code is committed I can bump up the submodule in GHC to kick the CI along.

snoyberg:
Cool, I'll wait a few days to hear from @bgamari but otherwise merge at the end of the week.

Mistuke:
Thanks!

PR #240, opened 2022-04-14
Support compiling for wasm32-wasi

TerrorJack:
This patch postpones compile-time errors to run-time, and allow Haskell projects that link to process but doesn't actually spawn child processes to build on wasm32-wasi successfully.

### DIFF
--- a/System/Process.hs
+++ b/System/Process.hs
@@ -77,6 +77,8 @@ module System.Process (
     rawSystem,
     ) where
 
+#include <ghcplatform.h>
+
 import Prelude hiding (mapM)
 
 import System.Process.Internals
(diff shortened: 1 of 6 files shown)


snoyberg:
I don't see an issue with this. Any thoughts @bgamari?

snoyberg:
This seems to have broken CI with:

```
Package check reported the following errors:
Unknown architecture name 'wasm32'
```

TerrorJack:
Yes, this is blocked on https://github.com/haskell/cabal/pull/8096 for the time being; if you wouldn't mind, let's leave this PR open for some more time, until `Cabal` tags a release with the fix, then it'll be possible to get CI green again on this PR.

TerrorJack:
@snoyberg I've changed this PR to avoid relying on `arch(wasm32)` Cabal conditional, so CI should be green, would you initiate another CI run and take a look?

snoyberg (commented):
It's certainly better that it doesn't break CI, but I'm still concerned about including code like this in a base library. @bgamari do you have any thoughts from the GHC side?

TerrorJack:
Closing in favor of #270.

PR #242, opened 2022-04-25
testsuite/process011: Don't rely on Python

### LINKED ISSUE #241: process011 is broken on Debian 10
It appears that the refactor of `process011` performed in 8cd7e043b8b55cca398ce4c3c3880aff3af62975, where `process011` is taught to use `python3` instead of `sh`, has broken the test on Debian 10. Specifically, for reasons that I don't yet understand `python3` exits with code 1 not code  2 when killing itself with signal 2:
```bash
$ (python3 -c 'import os; os.kill(os.getpid(), 2)' || echo $?)
Traceback (most recent call last):
  File "<string>", line 1, in <module>
KeyboardInterrupt
1
```

bgamari:
The refactor of `process011` performed in 8cd7e04, where
the test is taught to use `python3` instead of `sh`, has broken the test
on Debian 10. Specifically, for reasons that I don't yet understand,
`python3` exits with code 1 not code 2 when killing itself with
signal 2. Strangely, my tests on other distributions (e.g. NixOS)
suggest that Python's behavior in this case is rather inconsistent.

To avoid this, we now rather use a dedicated C program instead of
Python.

Fixes #241.

### DIFF
--- a/tests/process011.hs
+++ b/tests/process011.hs
@@ -14,24 +14,16 @@ main = do
 
   -- shell kills itself with SIGINT,
   -- delegation off, exit code (death by signal) reported as normal
-  do let script = intercalate " "
-                    [ "exec python3 2>/dev/null"
-                    , "-c"
-                    , "'import os; os.kill(os.getpid(), 2)'"
-                    ]
-     (_,_,_,p) <- createProcess (shell script) { delegate_ctlc = False }
+  do let script = "./process011_c"
+     (_,_,_,p) <- createProcess (proc script []) { delegate_ctlc = False }
      waitForProcess p >>= print
 
   putStrLn "===================== test 2"
 
   -- shell kills itself with SIGINT,
   -- delegation on, so expect to throw UserInterrupt
-  do let script = intercalate " "
-                    [ "exec python3 2>/dev/null"
-                    , "-c"
-                    , "'import os; os.kill(os.getpid(), 2)'"
-                    ]
-     (_,_,_,p) <- createProcess (shell script) { delegate_ctlc = True }
+  do let script = "./process011_c"
+     (_,_,_,p) <- createProcess (proc script []) { delegate_ctlc = True }
      (waitForProcess p >>= print)
        `catchUserInterrupt` \e -> putStrLn $ "caught: " ++ show e
 
--- a/tests/all.T
+++ b/tests/all.T
@@ -42,6 +42,8 @@ test('process010', [
     normalise_fun(lambda s: s.replace('illegal operation (Inappropriate ioctl for device)', 'does not exist (No such file or directory)')),
     normalise_exec
 ], compile_and_run, [''])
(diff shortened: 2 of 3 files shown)


bgamari:
I am validating this in https://gitlab.haskell.org/ghc/ghc/-/merge_requests/8014.

tests/process011.hs
```diff
@@ -14,23 +14,15 @@ main = do
 
   -- shell kills itself with SIGINT,
   -- delegation off, exit code (death by signal) reported as normal
-  do let script = intercalate " "
-                    [ "exec python3 2>/dev/null"
-                    , "-c"
-                    , "'import os; os.kill(os.getpid(), 2)'"
-                    ]
+  do let script = "./process011_c"
```
blackgnezdo, line 17: This should be `exec ./process011_c` to avoid the original issue of pdksh mangling the exit code. Alternatively an API which skips shell entirely should be used.

tests/process011.hs
```diff
@@ -14,23 +14,15 @@ main = do
...
   putStrLn "===================== test 2"
 
   -- shell kills itself with SIGINT,
   -- delegation on, so expect to throw UserInterrupt
-  do let script = intercalate " "
-                    [ "exec python3 2>/dev/null"
-                    , "-c"
-                    , "'import os; os.kill(os.getpid(), 2)'"
-                    ]
+  do let script = "./process011_c"
```
blackgnezdo, line 25: Likewise, `exec`.

blackgnezdo (commented):
With the noted `exec`s I confirmed that your change works for OpenBSD.

blackgnezdo:
Better yet, instead of `exec`, apply this patch:

```patch
diff --git a/tests/process011.hs b/tests/process011.hs
index 83e2030..b711fe4 100644
--- a/tests/process011.hs
+++ b/tests/process011.hs
@@ -15,7 +15,7 @@ main = do
   -- shell kills itself with SIGINT,
   -- delegation off, exit code (death by signal) reported as normal
   do let script = "./process011_c"
-     (_,_,_,p) <- createProcess (shell script) { delegate_ctlc = False }
+     (_,_,_,p) <- createProcess (proc script []) { delegate_ctlc = False }
      waitForProcess p >>= print
 
   putStrLn "===================== test 2"
@@ -23,7 +23,7 @@ main = do
   -- shell kills itself with SIGINT,
   -- delegation on, so expect to throw UserInterrupt
   do let script = "./process011_c"
-     (_,_,_,p) <- createProcess (shell script) { delegate_ctlc = True }
+     (_,_,_,p) <- createProcess (proc script []) { delegate_ctlc = True }
      (waitForProcess p >>= print)
        `catchUserInterrupt` \e -> putStrLn $ "caught: " ++ show e
 
```

bgamari:
Thanks @blackgnezdo !

PR #232, opened 2022-02-08
configure: Ensure that posix_spawnp reports ENOENT correctly

bgamari:
It turns out that POSIX allows implementations to report
errors "lazily" by exiting the spawned process with code 127.

In particular, OpenBSD's implementation returns successfully if the
requested executable doesn't exist. This severely hampers our ability to
report errors accurately.

In this patch we add a `configure` test for sensible error-reporting
from `posix_spawnp` and disable it if the test fails.

Fixes #224.

PR #244, opened 2022-05-21
fix cross-compiling configure issue

kgardas:
Hello,
quite recent addition for testing posix_spawn by running actual code broke cross-compiling capability. Simple hack fixes that
by assuming posix_spawn is working fine in this case.
E.g. see #232


snoyberg:
@bgamari would you mind reviewing this?

bgamari:
Indeed this seems quite reasonable to me. Thanks @kgardas!

PR #245, opened 2022-05-23
winio: Fix Handle mode on inherited handles.

Mistuke:
There's a small bug in the modes that we create handles with when we make inheritable handles.

We create both the input and output handles with `FILE_FLAG_OVERLAPPED`.
This is OK as long as the program being called is also not using IOCP. For the majority of the program
usually called this is the case and so we never noticed the bug in testing.

However when calling an application which is also using overlapped IO *or* uses the kernel low level
async APIs then the spawned application will misbehave because the operation would unexpectedly
become asynchronous.  This is the case with any program using Microsoft VC++'s stdlibs to read pipes.

The fix is to only mark handles with `FILE_FLAG_OVERLAPPED` if they are *not* being inherited.  That is,
only make them asynchronous if we, the caller will be using them and not the callee.

Note that WINIO can handle synchronous handles as well, so even if some other part of the program
uses the synchronous handle it would still work correctly.

Fixes https://gitlab.haskell.org/ghc/ghc/-/issues/21610

/cc @bgamari

### DIFF
--- a/cbits/win32/runProcess.c
+++ b/cbits/win32/runProcess.c
@@ -142,9 +142,10 @@ mkNamedPipe (HANDLE* pHandleIn, BOOL isInheritableIn,
        bytes and the error ERROR_NO_DATA."[0]
 
        [0] https://devblogs.microsoft.com/oldnewthing/20110114-00/?p=11753  */
+    DWORD inAttr = isInheritableIn ? 0 : FILE_FLAG_OVERLAPPED;
     hTemporaryIn
       = CreateNamedPipeW (pipeName,
-                          PIPE_ACCESS_INBOUND | FILE_FLAG_OVERLAPPED | FILE_FLAG_FIRST_PIPE_INSTANCE,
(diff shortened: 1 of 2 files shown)


snoyberg:
LGTM, though I can't claim I have a full understanding. Can you add a changelog entry about this? And does this warrant an immediate release, or should it be bundled with upcoming releases?

Mistuke:
> LGTM, though I can't claim I have a full understanding. Can you add a changelog entry about this? 

Sure, will do!

> And does this warrant an immediate release, or should it be bundled with upcoming releases?

I think that probably depends on if @mpickering things it can be included in the 9.4 release, otherwise no rush.

Mistuke:
@snoyberg changelog added. looks like there no need for an immediate release, just merging should be enough and I'll be able to close the GHC ticket. thanks!

snoyberg (approved):
Thanks!

PR #250, opened 2022-06-14
cbits/fork-exec: Don't dup2 identical fds

bgamari:
Darwin violates POSIX by making
`dup2(x,x)`, which should be a no-op, error. Consequently, we must take
care not to `dup2` in such cases.

We had already made this change in the `posix_spawnp` codepath but I had
assumed that this *only* affected `posix_spawnp`, not the `dup2` system
call itself.

See also: #214

### DIFF
--- a/cbits/posix/fork_exec.c
+++ b/cbits/posix/fork_exec.c
@@ -69,8 +69,13 @@ setup_std_handle_fork(int fd,
         return 0;
 
     case STD_HANDLE_USE_FD:
-        if (dup2(b->use_fd,  fd) == -1) {
-            child_failed(pipe, "dup2");
+        // N.B. POSIX specifies that dup2(x,x) should be a no-op, but
+        // naturally Apple ignores this and rather fails in posix_spawn on Big
+        // Sur.
+        if (b->use_fd != fd) {
+            if (dup2(b->use_fd,  fd) == -1) {
+                child_failed(pipe, "dup2");
+            }
         }
         return 0;
 


psibi:
I compiled Stack using this patch, but I'm still able to reproduce the issue when running stack's integration test. Some relevant logs:

```
/Users/ec2-user/.stack/programs/x86_64-osx/ghc-9.2.3/bin/ghc-9.2.3: startProcess: dup2: invalid argument (Bad file descriptor)
Main.hs: Exited with exit code: ExitFailure 1
CallStack (from HasCallStack):
  error, called at /Users/ec2-user/stack/test/integration/lib/StackTest.hs:63:34 in main:StackTest
  stack, called at /Users/ec2-user/stack/test/integration/tests/111-custom-snapshot/Main.hs:4:8 in main:Main
```

Unfortunately I haven't been able to find a minimal reproducible example.

bgamari:
Hmm, this is quite mysterious. Are you certain that you are linking against the correct `process`?

psibi:
I mis-applied your patch, I rebased your changes against master branch and it indeed fixes the issue. Relevant logs:

```
End of log for 111-custom-snapshot
Successful tests:
- 111-custom-snapshot

No failures!
```

Although I'm still able to reproduce the `posix_spawnp` issue which I have described in the issue.

psibi (approved):
This fixes the dup2 issue that was being reproduced in the stack codebase.

snoyberg:
Thanks @bgamari!

PR #255, opened 2022-07-22
Use bsd_closefrom.c from openssh-portable to fix #189

### LINKED ISSUE #189: Speed up close_fds with the new close_range() Linux/FreeBSD syscall

thomasjm:
I'm trying to fix #189.

Closing file descriptors in a reliable and portable way is not simple. The best summary I found of the different approaches available is in this [StackOverflow answer](https://stackoverflow.com/a/918469/2659595).

Modern versions of `glibc` (>= 2.34) and/or the Linux kernel (>= 5.9) have added support for the `closefrom` and `close_range` functions, which allow us to performantly do what we want. A good solution will use these functions when available, and otherwise use fallbacks. Putting together the fallbacks is a little involved.

In this PR I added the BSD-licensed `bsd_closefrom.c` file from https://github.com/openssh/openssh-portable/ and use it to implement the FD-closing logic from this library.

I haven't tested this yet, I'm putting it up to see how it does in CI and solicit feedback on the approach. Thanks!


snoyberg (approved):
Seems reasonable to me. @bgamari would you mind looking as well? You're far more familiar with the C code in this repo.

thomasjm:
Note: this isn't linking against musl libc at the moment, I probably need to get the macros in order...

bgamari:
@thomasjm, what is the status of this? Does this now work against `musl`?  The implementation looks good on a cursory look although this does need to be rebased so I have held off on a thorough review.

One thing that I am admittedly worried about here is the potential for C symbol name clashes. In particular, `process`'s `cbits` exposes many symbols with extremely generic names; this is certainly not a new problem but the fact that we now use code from OpenSSH increases the chance that things will break in the wild. Ideally, Haskell libraries would prefix all exported C symbols with a semi-unique prefix (e.g. `hs_process`).

thomasjm:
I ended up going with a workaround where I just lowered the max file descriptors in my Kubernetes environments, so this PR isn't quite done.

IIRC I was still fumbling around with CMake options and having trouble getting this to build with `musl`.

I can help push it over the finish line if there's interest.

bgamari:
It does seem like a nice improvement; it would be great to get it upstream.

process.cabal
```diff
@@ -66,18 +71,21 @@ library
...
             cbits/posix/fork_exec.c
             cbits/posix/posix_spawn.c
             cbits/posix/find_executable.c
+        includes:
+            runProcess.h
+            cbits/posix/bsd_closefrom.h
+        install-includes:
+            runProcess.h
+            processFlags.h
+            cbits/posix/bsd_closefrom.h
```
bgamari, line 84: Do we really need to install this header?

bgamari (changes requested):

cbits/posix/runProcess.c
```diff
@@ -21,21 +24,25 @@
...
-        cache = sysconf(_SC_OPEN_MAX);
-        if (cache == -1) {
-            cache = 256;
+#define CLOSE_RANGE_SYSCALL_NUMBER 436
+
+void
+closefrom_excluding(int lowfd, int excludingFd) {
+    // Try using the close_range syscall, provided in Linux kernel >= 5.9.
+    // We do this directly because not all C libs provide a wrapper (like musl)
+    long ret = syscall(CLOSE_RANGE_SYSCALL_NUMBER, lowfd, excludingFd - 1, 0);
```
thomasjm, line 33: This line is a little dodgy due to the direct syscall. It should be good on Linux kernels because Linux has very strong guarantees about syscall stability. For other POSIX platforms we probably need to gate this behind a macro. Another option would be to not even try using the `close_range` syscall, and just loop from `lowFd` to `excludingFd`, assuming that `excludingFd` will be reasonably small.

FWIW there's a possibility that `close_range` will be added to `musl` in the future: https://inbox.vuxu.org/musl/20220810130311.dwk7zxwkz32igrdm@yandex.com/T/#u

It would help me move this along if I had an easy way to test `musl` builds. I have a somewhat convoluted method using Haskell.nix, but do you happen to know a good way @bgamari ?

process.cabal
thomasjm: Probably not, I'll try to remove it once everything else is working smoothly.

cbits/posix/runProcess.c
bgamari: I would likely just use GHC's CI infrastructure. That is, open a draft MR against [`ghc/ghc`](https://gitlab.haskell.org/ghc/ghc) bumping the `process` submodule to your branch. This will get tested against Alpine 3.12.

bgamari:
Do let me know when you would like a review on this, @thomasjm .

process.cabal
thomasjm: Removed now

thomasjm:
Thanks @bgamari, I think it's just about ready for review!

> Ideally, Haskell libraries would prefix all exported C symbols with a semi-unique prefix (e.g. hs_process).

I've done this in the latest commit, does that look like what you had in mind?

The GHC pipeline is passing at this point. I think what we could still use is some test(s) that file descriptors actually get closed properly in `fork_exec.c`.

I'm also wondering what happens on *BSD or other platforms where the syscall could potentially differ. Poking around in the FreeBSD source it looks like the arguments match Linux. I'm still scared of the direct syscall, although I really want this to work on static Musl builds. FWIW, I just pushed a commit to bring `bsd_closefrom.c` up to date with a couple upstream changes I noticed we were missing. One of those changes involves an actual call `close_range`, gated on `HAVE_CLOSE_RANGE`. We should probably be consistent about this: if we decide a `close_range` syscall is okay, we should replace that call in `bsd_closefrom.c` also.

process.cabal
```diff
@@ -66,15 +68,17 @@ library
     else
         c-sources:
             cbits/posix/runProcess.c
+            cbits/posix/bsd_closefrom.c
             cbits/posix/fork_exec.c
             cbits/posix/posix_spawn.c
             cbits/posix/find_executable.c
+        includes:
+            runProcess.h
+            cbits/posix/bsd_closefrom.h
```
bgamari, line 77: I think you want to shorten this to `bsd_closefrom.h` and add `cbits/posix` to `include-dirs`.

bgamari (changes requested):

process.cabal
thomasjm: Done, I just resubmitted this to test in the GHC pipelines

PR #257, opened 2022-08-04
Fix process#251

### LINKED ISSUE #251: MacOS: posix_spawnp: invalid argument (Bad file descriptor)
Using the process shipped with GHC 9.2.3 (and even the latest master), I have been able to reproduce a bug while running integration tests for Stack. Following points about it:

- Reproducible only on MacOS. They work fine on both Windows and Linux.
- It's trivial to reproduce it on Mac with the stack integration test suite. I tried to get a minimal example, but my attempts have been futile.

For reproducing it via Mac, use this branch:

- macos-posix-spawn-bug: https://github.com/commercialhaskell/stack/tree/macos-posix-spawn-bug
- Lots of tests are affected by it. The logs usually look like this:

```
/Users/ec2-user/.stack/programs/x86_64-osx/ghc-9.2.3/bin/ghc-pkg-9.2.3: startProcess: posix_spawnp: invalid argument (Bad file descriptor)
Main.hs: Exited with exit code: ExitFailure 1
CallStack (from HasCallStack):
  error, called at /Users/ec2-user/stack/test/integration/lib/StackTest.hs:63:34 in main:StackTest
  stack, called at /Users/ec2-user/stack/test/integration/tests/script-extra-dep/Main.hs:4:8 in main:Main

End of log for script-extra-dep
Failed tests:
- script-extra-dep - ExitFailure 1
```

A quick way to reproduce it doing something like this `stack-integration-test -m 111-custom-snapshot`.

Note that the above branch is sprinkled with lots of logs. It was done to help debugging. Let me know if you want me to remove it. I have also sprinkled some logs on my [process fork](https://github.com/psibi/process/commit/82982c60ee47ab9e84d0df8fec323a494cd44a42) and it indicates to me that the bug happens when creating the `ghc-pkg` process. Also the log indicates that I get the exit status code of 9.

I invoked dtruss using the following command:

```
sudo dtruss ~/.local/bin/stack-integration-test -m 111-custom-snapshot 
```

This is the dtruss logs for the above command: https://gist.github.com/psibi/0c3c89dd2b90012d7d9f3a64ceffb73a

bgamari:
Previously to spawn a process with a closed standard handle, we
would use `posix_spawn_file_action_addclose`. However, it turns out that
POSIX specifies that `spawnp()` may fail if `addclose()` is used on an
fd that is already closed. While glibc and musl appear to ignore this
aspect of the specification, Darwin indeed follows it leading to #251.

This behavior is rather unfortunate as
`posix_spawn_file_action_addclose` is a convenient way to close a handle
in a subprocess in a race-free manner (e.g. unlike `O_CLOEXEC`, which is
global). To avoid #251 we must first use
`posix_spawn_file_action_addopen` on the fd (e.g. opening `/dev/null`)
to be closed to ensure that it is valid, which has the side-effect of
closing the inherited fd. We can then safely use
`posix_spawn_file_action_addclose` to close the fd.

Fixes #251.

### DIFF
--- a/cbits/posix/posix_spawn.c
+++ b/cbits/posix/posix_spawn.c
@@ -4,6 +4,7 @@
 #include <unistd.h>
 #include <errno.h>
 #include <signal.h>
+#include <fcntl.h>
 
 #if !defined(USE_POSIX_SPAWN)
 ProcHandle
@@ -35,6 +36,14 @@ setup_std_handle_spawn (int fd,
 {
     switch (hdl->behavior) {
     case STD_HANDLE_CLOSE:
+        // N.B. POSIX specifies that addclose() may result in spawnp() failing
+        // if the fd to-be-closed is already closed. Consequently, we must
+        // first open a file (e.g. /dev/null) and before attempting to close
+        // the fd. Fixes #251.
+        if (posix_spawn_file_actions_addopen(fa, fd, "/dev/null", O_RDONLY, 0) != 0) {
+            *failed_doing = "posix_spawn_file_actions_addopen";
+            return -1;
+        }
         if (posix_spawn_file_actions_addclose(fa, fd) != 0) {
             *failed_doing = "posix_spawn_file_actions_addclose";
             return -1;
--- a/tests/all.T
+++ b/tests/all.T
@@ -47,3 +47,4 @@ test('process011',
      compile_and_run, [''])
 
 test('T8343', normal, compile_and_run, [''])
+test('processT251', normal, compile_and_run, [''])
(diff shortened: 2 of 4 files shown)


bgamari:
Unfortunately I found that the current Cabal-driven testsuite isn't particularly well-suited to test this, consequently I ended up adding a test in GHC's testsuite. Testing this locally as we speak.

bgamari:
I have confirmed that the test reproduces #251 on Darwin without this change and that the change fixes it.

bgamari:
Note that I ultimately opted not to make the `addopen`/`addclose` behavior conditional on platform since technically `spawnp` failing in this case is what is mandated by POSIX.

hasufell:
Is this released? I can't find it in https://hackage.haskell.org/package/process-1.6.15.0/changelog

PR #270, opened 2023-01-04
Fix compilation for wasm32-wasi

TerrorJack:
This one is a rework of #240 and shouldn't break CI in this repo or any other GHC platforms now.

### DIFF
--- a/System/Process.hs
+++ b/System/Process.hs
@@ -102,6 +102,11 @@ import System.Posix.Types (CPid (..))
 
 import GHC.IO.Exception ( ioException, IOErrorType(..), IOException(..) )
 
+#if defined(__wasm__)
+import GHC.IO.Exception ( unsupportedOperation )
+import System.IO.Error
+#endif
+
 -- | The platform specific type for a process identifier.
 --
 -- This is always an integral type. Width and signedness are platform specific.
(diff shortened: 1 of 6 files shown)


TerrorJack:
@snoyberg Would you allow the workflow to run on this PR? Thanks :)

snoyberg:
PR is passing CI, and I don't see any issues with this. @bgamari would you mind confirming that you're OK with this change landing?

TerrorJack:
Ben is on holiday yet; also allow me to draft it atm, there were spurious "redundant import" warnings that turned to errors when testing in non-wasm jobs, so I also need a bit of time to look into it. Will undraft and ping again once I'm absolutely sure it doesn't break stuff on other platforms :/

TerrorJack:
Ping; the current patch is good to go :)

System/Process.hs
```diff
@@ -857,6 +862,19 @@ terminateProcess ph = do
 -- ----------------------------------------------------------------------------
 -- Interface to C bits
 
+#if defined(__wasm__)
+
+c_terminateProcess :: PHANDLE -> IO CInt
+c_terminateProcess _ = ioError (ioeSetLocation unsupportedOperation "terminateProcess")
```
bgamari, line 868: Yes, this is better than the previous iteration. I probably would have factored out this pattern into a `unsupportedByWasm` helper function living in `System.Process.Internal` but I'm fine with this as well.

bgamari (approved):

PR #271, opened 2023-01-16
Fix a non-existent macro which is unset by GHC

TerrorJack:
This PR corrects an unfortunate mistake in #270. The `__wasm__` macro is a builtin system macro defined by `clang` when targetting `wasm32`, however, the autoconf scripts in GHC will automatically add `-undef` to the CPP flags that GHC uses to invoke `clang`, therefore this flag is unset and will result in a link-time error for `wasm32-wasi-ghc`.

After this fix lands, the existing test for cross GHC will also be updated to ensure these `process` functions are included as a part of the test to avoid any regression here.

### DIFF
--- a/System/Process.hs
+++ b/System/Process.hs
@@ -6,6 +6,8 @@
 #endif
 {-# LANGUAGE InterruptibleFFI #-}
 
+#include <ghcplatform.h>
+
 -----------------------------------------------------------------------------
 -- |
 -- Module      :  System.Process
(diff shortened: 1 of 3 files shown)


TerrorJack:
@snoyberg It seems the pipeline isn't running even though I've got PRs merged before, would you be able to start it? Thanks :)

snoyberg:
It's not giving me an option to approve starting CI, I'm not sure why.

TerrorJack:
Closing in favor of #272.

PR #272, opened 2023-01-17
Fix a non-existent macro unset by GHC

TerrorJack:
This PR corrects an unfortunate mistake in #270. The `__wasm__` macro is a builtin system macro defined by `clang` when targetting `wasm32`, however, the autoconf scripts in GHC will automatically add `-undef` to the CPP flags that GHC uses to invoke `clang`, therefore this flag is unset and will result in a link-time error for `wasm32-wasi-ghc`.

After this fix lands, the existing test for cross GHC will also be updated to ensure these `process` functions are included as a part of the test to avoid any regression here.

### DIFF
--- a/System/Process.hs
+++ b/System/Process.hs
@@ -6,6 +6,8 @@
 #endif
 {-# LANGUAGE InterruptibleFFI #-}
 
+#include <ghcplatform.h>
+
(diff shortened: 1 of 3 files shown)


TerrorJack:
This PR's pipeline is passing. 👀

PR #277, opened 2023-03-05
Windows: don't wait indefinitely on processes

### LINKED ISSUE #273: on Windows, `waitForProcess` hangs forever on certain `git` commands
Encountered in this cabal issue: https://github.com/haskell/cabal/issues/8688

Running the following script will successfully execute the first process, but then will get stuck indefinitely while waiting for the second process to finish. The program cannot be killed with <kbd>Ctr</kbd>+<kbd>C</kbd>, but can be killed from the task manager.

If the process is killed and the `git fetch` command is run manually, the program will succeed if it's run again.

In WSL, the script ran successfully.

### Reproduction steps:

Run the following script with `cabal run <filename>`:

```haskell
{- cabal:
build-depends: base >= 4.16 && < 5
             , process == 1.6.16.0
-}
import Data.Foldable
import System.Process

processes :: [CreateProcess]
processes =
  [ CreateProcess {cmdspec = RawCommand "git" ["clone","--no-checkout","https://github.com/rtyley/small-test-repo"], cwd = Nothing, env = Nothing, std_in = Inherit, std_out = Inherit, std_err = Inherit, close_fds = False, create_group = False, delegate_ctlc = True, detach_console = False, create_new_console = False, new_session = False, child_group = Nothing, child_user = Nothing, use_process_jobs = True}
  , CreateProcess {cmdspec = RawCommand "git" ["fetch"], cwd = Just "./small-test-repo", env = Nothing, std_in = Inherit, std_out = Inherit, std_err = Inherit, close_fds = False, create_group = False, delegate_ctlc = True, detach_console = False, create_new_console = False, new_session = False, child_group = Nothing, child_user = Nothing, use_process_jobs = True}
  ]

main :: IO ()
main = do
  for_ processes $ \cp -> withCreateProcess cp $ \_ _ _ p -> do
    waitForProcess p >>= print
```

### Environment:

- Windows 11
- ghc 9.4.4
- git version 2.38.1.windows.1
- process 1.6.16.0

Mistuke:
The Windows kernel will release a process only when the last handle to it is closed.

This means that the call to `WaitForSingleObject` does not return until every other process closes any handles they may have opened to the program, including pipes passed to it.

However if on the Haskell side we haven't released the handle yet then we deadlock and wait indefinitely.  There is a race condition in that if when we get to `WaitForSingleObject` the kernel has already cleaned up the process then we'll return immediately, which is what is happening with the first test in the test program.

The old workflow for this wait function used to be that we iterate until the process list is empty.  On each we wait indefinitely.

The new workflow is different in that instead of waiting indefinitely we wait for 200ms and re-test the application.  Essentially we poll the first process that doesn't return `STILL_ACTIVE`.   It's less efficient than before but it's the only way to avoid the race condition.

Fixes #273 
Fixes https://github.com/haskell/cabal/issues/8688
Fixes https://github.com/haskell/cabal/issues/8208

Waiting for GHC bootstrap and testsuite to finish https://gitlab.haskell.org/ghc/ghc/-/merge_requests/10080?diff_id=193677

/CC @bgamari @snoyberg

### DIFF
--- a/cbits/win32/runProcess.c
+++ b/cbits/win32/runProcess.c
@@ -617,7 +617,22 @@ waitForJobCompletion ( HANDLE hJob )
         return true;
       }
 
-      HANDLE pHwnd = OpenProcess(SYNCHRONIZE, TRUE, pid_list->ProcessIdList[0]);
+      HANDLE pHwnd;
+      bool found_p;
+
+      /* Find the first non-terminated handle.  */
+      for (DWORD i = 0; i < pid_list->NumberOfProcessIdsInList; i++) {
+        pHwnd = OpenProcess(SYNCHRONIZE, FALSE, pid_list->ProcessIdList[i]);
+        int code = 0;
+        found_p = pHwnd && getProcessExitCode (pHwnd, &code) != -1 && code != STILL_ACTIVE;
+        if (found_p)
+          break;
+      }
+
+      /* All processes were terminated already.  */
+      if (!found_p)
+        return true;
+
       if (pHwnd == NULL) {
         switch (GetLastError()) {
           case ERROR_INVALID_PARAMETER:
@@ -632,7 +647,13 @@ waitForJobCompletion ( HANDLE hJob )
       }
 
       // Wait for it to finish...
-      if (WaitForSingleObject(pHwnd, INFINITE) != WAIT_OBJECT_0) {
+      DWORD result = WaitForSingleObject(pHwnd, 200);
+      if (result == WAIT_TIMEOUT) {
+        int code = 0;
+        /* Current process is waiting to terminate. skip.  */
+        if (getProcessExitCode (pHwnd, &code) != -1 && code != STILL_ACTIVE)
+          continue ;
+      } else if (result != WAIT_OBJECT_0) {
         free(pid_list);
         maperrno();
         CloseHandle(pHwnd);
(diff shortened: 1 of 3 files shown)


cbits/win32/runProcess.c
```diff
@@ -617,7 +617,22 @@ waitForJobCompletion ( HANDLE hJob )
         return true;
       }
 
-      HANDLE pHwnd = OpenProcess(SYNCHRONIZE, TRUE, pid_list->ProcessIdList[0]);
+      HANDLE pHwnd;
+      bool found_p;
+
+      /* Find the first non-terminated handle.  */
+      for (DWORD i = 0; i < pid_list->NumberOfProcessIdsInList; i++) {
+        pHwnd = OpenProcess(SYNCHRONIZE, FALSE, pid_list->ProcessIdList[0]);
```
robx, line 625: shouldn't this be `pid_list->ProcessIdList[i]`?

Mistuke: oops, indeed

snoyberg (changes requested):
LGTM. One request: could you add an explanation of the change to the changelog?

Mistuke:
Added a changelog entry

snoyberg (approved):

PR #278, opened 2023-03-11
Windows: Implement delegate_ctlc on Windows

Mistuke:
This implements `delegate_ctlc` on Windows so the calling process can properly ignore SIGINT.

Without this cabal repl doesn't behave correctly.

validated working with manual tests and cabal.

to address https://github.com/haskell/cabal/issues/8820


Mistuke:
Ready for review,

/cc @snoyberg

snoyberg (approved):

PR #280, opened 2023-03-13
cbits: Ensure that fork pipe doesn't shadow std fds

### LINKED ISSUE #266: processT251 failure (fd 0 considered special?)
When running on OpenBSD 7.2 as a part of my attempt to test ghc HEAD I ran into a failure which is easier to observe standalone:
```
% ghc --version
The Glorious Glasgow Haskell Compilation System, version 9.2.4
% ghc ~/s/ghc/libraries/process/tests/processT251.hs -o /tmp/z
ghc ~/s/ghc/libraries/process/tests/processT251.hs -o /tmp/z
[1 of 1] Compiling Main             ( /home/greg/s/ghc/libraries/process/tests/processT251.hs, /home/greg/s/ghc/libraries/process/tests/processT251.o )
Linking /tmp/z ...
% /tmp/z
parent start
child start
processT251: ./processT251: createProcess: close: invalid argument (Bad file descriptor)
z: user error (Pattern match failure in 'do' block at /home/greg/s/ghc/libraries/process/tests/processT251.hs:20:5-15)
% 
```
Tracing the syscalls is a bit tedious but boils down to these salient parts:
```
 46520 processT251 GIO   fd 1 wrote 13 bytes
       "parent start
       "
 46520 processT251 CALL  pipe(0x7f7ffffca5e8)
 46520 processT251 STRU  int [2] { 3, 4 }
 46520 processT251 RET   pipe 0
 46520 processT251 CALL  vfork()
 89659 processT251 RET   vfork 0
 89659 processT251 CALL  execve(0x224e040f0,0x224e04158,0x7f7ffffce9a8)
 89659 processT251 NAMI  "./processT251"
 89659 processT251 ARGS  
	[0] = "./processT251"
	[1] = "child"
 46520 processT251 RET   vfork 89659/0x15e3b
...
 89659 processT251 GIO   fd 1 wrote 12 bytes
       "child start
       "
 89659 processT251 RET   write 12/0xc
 89659 processT251 CALL  pipe(0x7f7ffffec6f8)
 89659 processT251 STRU  int [2] { 0, 3 }      ---------------------------------- [1]
 89659 processT251 RET   pipe 0
 89659 processT251 CALL  vfork()

 23176 processT251 RET   vfork 0
 23176 processT251 CALL  close(0)               ----------------------------------[2]
 23176 processT251 RET   close 0
 23176 processT251 CALL  fcntl(3,F_SETFD,FD_CLOEXEC)
 23176 processT251 RET   fcntl 0
...
 23176 processT251 CALL  close(0)              -----------------------------------[3]
 23176 processT251 RET   close -1 errno 9 Bad file descriptor
 23176 processT251 CALL  write(3,0x7f7ffffec628,0x8)
 23176 processT251 GIO   fd 3 wrote 8 bytes
       "\M-h\M^Z \0\0\0\0\0"
 23176 processT251 RET   write 8
 23176 processT251 CALL  write(3,0x7f7ffffec634,0x4)
 23176 processT251 GIO   fd 3 wrote 4 bytes
       "	\0\0\0"
```

So the "child" is PID 89659. 
 [1] "child" gets a pipe with a pair of descriptors [0, 3] 
 [2] "child2" closes fd0 successfully once
 [3] "child2" fails to close fd0 the second time (it's no longer open)

Maybe there's an expectation somewhere that `pipe` won't allocate these FDs because they are reserved? I don't think they are per any standard and since "child" already wanted to run with `std_in = NoStream`, its wish is fulfilled and fd=0 is used otherwise.

bgamari:
Previously we would assume that the pipe used to communicate errors from the child back to the parent did not shadow the standard descriptors (stdin, stdout, and stderr). Somewhat surprisingly, this appears to hold most platforms. However, OpenBSD appears to be a notable exception.

Avoid relying on this assumption by `dup`ing the pipe fds until they end up out of the standard fd range.

Closes #266.

### DIFF
--- a/cbits/posix/fork_exec.c
+++ b/cbits/posix/fork_exec.c
@@ -102,6 +102,32 @@ setup_std_handle_fork(int fd,
     }
 }
 
+/* We must ensure that the fork communications pipe does not inhabit fds 0
+ * through 2 since we will need to manipulate these fds in
+ * setup_std_handle_fork while keeping the pipe available so that it can report
+ * errors. See #266.
+ */
+int unshadow_pipe_fd(int fd, char **failed_doing) {
+    if (fd <= 2) {
+        int fd2 = dup(fd);
+        if (fd2 == -1) {
+            *failed_doing = "dup(unshadow)";
+            return -1;
+        }
+
+        // This should recurse at most three times
+        int fd3 = unshadow_pipe_fd(fd2, failed_doing);
+        if (close(fd2) == -1) {
+            *failed_doing = "close(unshadow)";
+            return -1;
+        }
+
+        return fd3;
+    } else {
+        return fd;
+    }
+}
+
 /* Try spawning with fork. */
 ProcHandle
 do_spawn_fork (char *const args[],
@@ -117,6 +143,16 @@ do_spawn_fork (char *const args[],
     int r = pipe(forkCommunicationFds);
     if (r == -1) {
         *failed_doing = "pipe";
+        return -1;
+    }
+
+    // Ensure that the pipe fds don't shadow stdin/stdout/stderr
+    forkCommunicationFds[0] = unshadow_pipe_fd(forkCommunicationFds[0], failed_doing);
+    if (forkCommunicationFds[0] == -1) {
+        return -1;
+    }
+    forkCommunicationFds[1] = unshadow_pipe_fd(forkCommunicationFds[1], failed_doing);
+    if (forkCommunicationFds[1] == -1) {
         return -1;
     }
 


cbits/posix/fork_exec.c
```diff
@@ -102,6 +102,32 @@ setup_std_handle_fork(int fd,
...
 }
 
+/* We must ensure that the fork communications pipe does not inhabit fds 0
+ * through 2 since we will need to manipulate these fds in
+ * setup_std_handle_fork while keeping the pipe available so that it can report
+ * errors. See #266.
+ */
+int unshadow_pipe_fd(int fd, char **failed_doing) {
+    if (fd <= 2) {
+        int fd2 = dup(fd);
```
blackgnezdo, line 112: I tested this, it seems to not be doing what's expected. The trace looks like this:
```
 54688 processT251 STRU  int [2] { 0, 3 }
 54688 processT251 RET   pipe 0
...
 54688 processT251 CALL  dup(0)
 54688 processT251 RET   dup 4
...
 54688 processT251 CALL  close(4)
 54688 processT251 RET   close 0
```

The recursive call below returns the same same 4 that it was passed in (i.e. both fd3=fd2=4) which then gets closed before it's returned, so the caller gets an invalid fd and the test reports:
```
$ ./processT251 
parent start
child start
processT251: ./processT251: createProcess: read pipe: invalid argument (Bad file descriptor)
processT251: user error (Pattern match failure in 'do' block at processT251.hs:20:5-15)
child2 start
child2 done
```

blackgnezdo: I'm testing this now:
```patch
diff --git a/cbits/posix/fork_exec.c b/cbits/posix/fork_exec.c
index 9c98446..a613763 100644
--- a/cbits/posix/fork_exec.c
+++ b/cbits/posix/fork_exec.c
@@ -108,24 +108,23 @@ setup_std_handle_fork(int fd,
  * errors. See #266.
  */
 int unshadow_pipe_fd(int fd, char **failed_doing) {
-    if (fd <= 2) {
-        int fd2 = dup(fd);
-        if (fd2 == -1) {
-            *failed_doing = "dup(unshadow)";
-            return -1;
-        }
-
-        // This should recurse at most three times
-        int fd3 = unshadow_pipe_fd(fd2, failed_doing);
-        if (close(fd2) == -1) {
-            *failed_doing = "close(unshadow)";
-            return -1;
-        }
-
-        return fd3;
-    } else {
-        return fd;
-    }
+	int i = 0;
+	int fds[3] = {0};
+	for (i = 0; fd < 3 && i < 3; ++i) {
+		fds[i] = fd;
+		fd = dup(fd);
+		if (fd == -1) {
+			*failed_doing = "dup(unshadow)";
+			return -1;
+		}
+	}
+	for (int j = 0; j < i; ++j) {
+		if (close(fds[j]) == -1) {
+			*failed_doing = "close(unshadow)";
+			return -1;
+		}
+	}
+	return fd;
 }
 
 /* Try spawning with fork. */
```

bgamari: Any update here, @blackgnezdo ?

blackgnezdo: I kicked the tires with a hand-written test program and it works. For some reason I can't make the test pass when running from within the ghc test suite. The trace looks different though after my change

```
 19355 processT251 STRU  int [2] { 0, 3 }
 19355 processT251 RET   pipe 0
 19355 processT251 CALL  dup(0)
 19355 processT251 RET   dup 4
 19355 processT251 CALL  close(0)                  <- closes the "bad" descriptor now
 19355 processT251 RET   close 0
 19355 processT251 CALL  vfork()
 90064 processT251 RET   vfork 0
 90064 processT251 CALL  close(4)
 90064 processT251 RET   close 0
 90064 processT251 CALL  fcntl(3,F_SETFD,FD_CLOEXEC)
 90064 processT251 RET   fcntl 0
 90064 processT251 CALL  sigprocmask(SIG_SETMASK,0<>)
 90064 processT251 RET   sigprocmask 0x2<SIGINT>
 90064 processT251 CALL  close(0)
 90064 processT251 RET   close -1 errno 9 Bad file descriptor
```

Looks like something else is broken. I'd debug it, but I need to figure out how to build the test program with a modified process library without rebuilding all of ghc.

blackgnezdo: I stared at this a bit more. I attribute the failure of close above to 
```
setup_std_handle_fork(STDIN_FILENO,  stdInHdl,  forkCommunicationFds[1]);
```
which tries to close the `fd` which is already closed. The test in question finally passes when I make close failure non-fatal:
```diff
diff --git a/cbits/posix/fork_exec.c b/cbits/posix/fork_exec.c                                                                                                                                
index a613763..82cc7af 100644                                                                                                                                                                 
--- a/cbits/posix/fork_exec.c                                                                                                                                                                 
+++ b/cbits/posix/fork_exec.c                                                                                                                                                                 
@@ -65,7 +65,7 @@ setup_std_handle_fork(int fd,                                                                                                                                               
 {                                                                                                                                                                                            
     switch (b->behavior) {                                                                                                                                                                   
     case STD_HANDLE_CLOSE:                                                                                                                                                                   
-        if (close(fd) == -1) {                                                                                                                                                               
+        if (close(fd) == -1 && errno != EBADF) {                                                                                                                                             
             child_failed(pipe, "close");                                                                                                                                                     
         }                                                                                                                                                                                    
         return 0;                                                                                                                                                                            


```

blackgnezdo: #282 has a solution

bgamari:
This has been superceded by #282.

PR #282, opened 2023-03-21
cbits: Ensure that fork pipe doesn't shadow std fds 

blackgnezdo:
@bgamari, this makes `processT251` pass on OpenBSD


bgamari:
Thanks @blackgnezdo !

Issue #284, opened 2023-03-30
Use `O_DUPFD` instead of `dup`


bgamari:
In #282 we introduced some rather subtle `dup` logic to ensure that the fork pipe fds don't shadow stdin/stdout/stderr. However, it turns out that POSIX provides a much simpler means of doing this: `fcntl(fd, O_DUPFD)` ([manpage](https://pubs.opengroup.org/onlinepubs/9699919799/functions/fcntl.html#)). We should rather use this.

bgamari:
Yes, fair point.

PR #285, opened 2023-03-30
Address #284 and #283

### LINKED ISSUE #283: Failing `find_executable` results in timer remaining stopped
Before we call `find_executable` we call `stopTimer` to avoid `fork` being interrupted by a signal:

https://github.com/haskell/process/blob/c956562fe55bca1465c7a9808ab9b293925dd2bc/cbits/posix/fork_exec.c#L166

However, in the event that `find_executable` fails we do not restart the timer:

https://github.com/haskell/process/blob/c956562fe55bca1465c7a9808ab9b293925dd2bc/cbits/posix/fork_exec.c#L175-L180

I believe the solution to this is to simply move the `stopTimer` call after `find_executable`.

bgamari:
Fix #283, ensuring that failure of `find_executable` doesn't permanently stop the GHC ticker.

Address #284 by implementing `unshadow_pipe_fd` using `fcntl(F_DUPFD)`, considerably simplifying its implementation.

### DIFF
--- a/cbits/posix/fork_exec.c
+++ b/cbits/posix/fork_exec.c
@@ -107,23 +107,17 @@ setup_std_handle_fork(int fd,
  * errors. See #266.
  */
 int unshadow_pipe_fd(int fd, char **failed_doing) {
-	int i = 0;
-	int fds[3] = {0};
-	for (i = 0; fd < 3 && i < 3; ++i) {
-		fds[i] = fd;
-		fd = dup(fd);
-		if (fd == -1) {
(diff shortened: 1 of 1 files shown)


bgamari:
@blackgnezdo, perhaps you could have a look at this?

blackgnezdo:
> @blackgnezdo, perhaps you could have a look at this?

Works like a charm. Thanks for simplifying it so nicely.

snoyberg (approved):

PR #288, opened 2023-04-10
cbits/fork_exec: Eliminate potential leakage of comm. pipes

### LINKED ISSUE #287: fork/exec implementation may leak fds in concurrent program
Currently the `fork`/`exec` spawn implementation does not use `O_CLOEXEC` when available. This means that the communications pipe of one thread's `createProcess` may leak into another thread's child process. This could result in strange failures (e.g. if an errant child writes to the pipe) or, in principle, running out of fds (e.g. if a thread somehow races with thousands of concurrent spawns).

bgamari:
Previously we used `fork` followed by `fcntl` to create a pipe with `O_CLOEXEC`. However, this is potentially racy as another thread may `fork` in between in these two calls, resulting in the pipe of the former thread being available to the child of the latter.

Fix this by using `O_CLOEXEC` when creating the communications pipe.

Closes #287.

### DIFF
--- a/cbits/posix/fork_exec.c
+++ b/cbits/posix/fork_exec.c
@@ -31,10 +31,7 @@
 
 #include <Rts.h>
 
-#if defined(HAVE_WORKING_FORK)
-#define myfork fork
-// We don't need a fork command on Windows
-#else
+#if !defined(HAVE_WORKING_FORK)
 #error Cannot find a working fork command
 #endif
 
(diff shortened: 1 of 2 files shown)


bgamari:
@expipiplus1, perhaps you could have a look at this? This should address your concern from https://github.com/haskell/process/issues/284#issuecomment-1491964449.

snoyberg (approved):

expipiplus1:
@bgamari yes, I think so (assuming that the unshadow function is only used for pipes which should be closed, which I believe it currently is)

PR #296, opened 2023-07-25
Prevent a segfault in createProcess on Mac

### LINKED ISSUE #295: createProcess segfaults on Mac
The following test case produces a segfault with `ghc-9.4.5` on Mac. The version of `process` is `1.6.16.0`.

```haskell
import Control.Monad
import System.Process

main :: IO ()
main = void $ createProcess $ CreateProcess
    { child_group = Nothing
    , child_user = Nothing
    , close_fds = False
    , cmdspec = RawCommand "env" []
    , create_group = True
    , create_new_console = False
    , cwd = Nothing
    , delegate_ctlc = False
    , detach_console = False
    , env = Just [("PATH", "/bin:/usr/bin")]
    , new_session = False
    , std_err = Inherit
    , std_in = Inherit
    , std_out = Inherit
    , use_process_jobs = False
    }
```

I used the copy of `ghc` installed by `stack`:

```shellsession
$ ~/.stack/programs/x86_64-osx/ghc-9.4.5/bin/runghc TestCase.hs
Segmentation fault: 11
$ ~/.stack/programs/x86_64-osx/ghc-9.4.5/bin/ghc-pkg list process
~/.stack/programs/x86_64-osx/ghc-9.4.5/lib/ghc-9.4.5/lib/package.conf.d
    process-1.6.16.0
```

It's crashing in `strlen`:

```shellsession
$ lldb TestCase
(lldb) target create "TestCase"
Current executable set to '~/portal-suite/devtools/gabe/test/TestCase' (x86_64).
(lldb) r
Process 65154 launched: '~/portal-suite/devtools/gabe/test/TestCase' (x86_64)
Process 65154 stopped
* thread #1, name = 'ghc_ticker', queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
    frame #0: 0x00007ff8189646b2 libsystem_platform.dylib`_platform_strlen + 18
libsystem_platform.dylib`:
->  0x7ff8189646b2 <+18>: pcmpeqb (%rdi), %xmm0
    0x7ff8189646b6 <+22>: pmovmskb %xmm0, %esi
    0x7ff8189646ba <+26>: andq   $0xf, %rcx
    0x7ff8189646be <+30>: orq    $-0x1, %rax
Target 0: (TestCase) stopped.
(lldb) p $rdi
(unsigned long) $0 = 0
(lldb) bt
* thread #1, name = 'ghc_ticker', queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
  * frame #0: 0x00007ff8189646b2 libsystem_platform.dylib`_platform_strlen + 18
    frame #1: 0x000000010000d0ae TestCase`find_executable + 142
    frame #2: 0x000000010000c873 TestCase`do_spawn_fork + 99
    frame #3: 0x000000010000c6bd TestCase`runInteractiveProcess + 429
    frame #4: 0x0000000100010ce1 TestCase`Lc5h0_info + 113
```

Although the backtrace says it's in `find_executable`, I think it's actually in `find_in_search_path` which has been inlined by the C compiler:

```c
static char *find_in_search_path(char *working_dir, char *search_path, const char *filename) {
    int workdir_len = strlen(working_dir);
```

It doesn't crash when the Haskell `cwd` value is a `Just`.

neilmayhew:
The new test segfaults without the corresponding change in `find_in_search_path`

Using `.` is slightly less efficient than not using a working directory at all, but it avoids adding a third code branch and the consequent additional duplication of `sprintf` code.

Closes #295

### DIFF
--- a/cbits/posix/find_executable.c
+++ b/cbits/posix/find_executable.c
@@ -53,10 +53,12 @@ static bool is_executable(char *working_dir, const char *path) {
  * found.
   */
 static char *find_in_search_path(char *working_dir, char *search_path, const char *filename) {
-    int workdir_len = strlen(working_dir);
     const int filename_len = strlen(filename);
     char *tokbuf;
     char *path = strtok_r(search_path, ":", &tokbuf);
+    if (!working_dir)
+        working_dir = ".";
+    int workdir_len = strlen(working_dir);
     while (path != NULL) {
 	// N.B. gcc 6.3.0, used by Debian 9, inexplicably warns that `path`
 	// may not be initialised with -Wall.  Silence this warning. See #210.
--- a/test/main.hs
+++ b/test/main.hs
@@ -41,6 +41,7 @@ main = do
     testInterruptWith
     testDoubleWait
     testKillDoubleWait
+    testCreateProcess
     putStrLn ">>> Tests passed successfully"
 
 run :: String -> IO () -> IO ()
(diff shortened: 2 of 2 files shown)


neilmayhew:
I tested this on Mac and Linux as follows:

```shell
nix-shell -p haskell.compiler.ghc945 autoconf
autoreconf
ghc Setup.hs
./Setup configure --enable-tests
./Setup build
./Setup test
```

bgamari:
Thank you, @neilmayhew ! This looks good.

snoyberg (approved):

neilmayhew:
Thanks for approving and merging.

Since this is a boot package, I assume it will be a while before the fix gets into a version that we can use in an application.

Is there any chance this would be backported into a version of ghc 9.4? Or is there now a way to use a non-default version of a boot package?

Bodigrim:
> Or is there now a way to use a non-default version of a boot package?

It depends on your other dependencies, but with process you can likely say `cabal build --constraint 'process >= X.Y.Z'` and Cabal will rebuild everything in the world. 

GHC 9.4.6 is in works right now, you might want to chime in https://gitlab.haskell.org/ghc/ghc/-/issues/23760 and https://gitlab.haskell.org/ghc/ghc/-/issues/23710

andreasabel:
I chimed in on https://gitlab.haskell.org/ghc/ghc/-/issues/23760 because not even GHCs 9.4.7 and 9.6.3 contain this important bugfix.

PR #297, opened 2023-11-27
Remove tests/ subdirectory

BinderDavid:
The tests in the `tests` subdirectory are only run by the GHC testsuite. Following https://gitlab.haskell.org/ghc/ghc/-/issues/22622 the tests will be moved from the git submodule to the `testsuite/tests` directory in GHC itself.


BinderDavid:
I'm still working on the patch on the ghc side for this which will add all the tests directly to the GHC testsuite.

bgamari:
Thanks @BinderDavid !

PR #300, opened 2024-01-18
Bump Win32

tomjaguarpaw:


tomjaguarpaw:
I'd like to check that process works with the new version of Win32 but I'm not sure that this CI can test that, because it uses a stack resolver rather than the latest version of Win32 on Hackage. I don't have a Windows machine myself. Does anyone have any clever ideas about how I can test this? (I guess I could just add a cabal-resolved job to the build matrix.)

bgamari:
I have validated this while bumping GHC's `process` submodule.

tomjaguarpaw:
Thanks, I've made a revision: https://hackage.haskell.org/package/process-1.6.18.0/revisions/

Issue #303, opened 2024-02-08
Dropping vfork support regressed performance when posix_spawn not available


adamgundry:
#279 replaced `vfork` calls with `fork`, but `fork` can perform substantially worse when the process has a large memory footprint.

Ideally we would use `posix_spawn` instead, but (when setting the cwd) this requires `posix_spawn_file_actions_addchdir`. There seems to be an inconsistency as the [`configure` script looks for `posix_spawn_file_actions_addchdir`](https://github.com/haskell/process/blob/dfdae0a7036b42d352a515214e6116424dd08ec9/configure.ac#L26) but [`do_spawn_posix` tests for the `_np` version](https://github.com/haskell/process/blob/dfdae0a7036b42d352a515214e6116424dd08ec9/cbits/posix/posix_spawn.c#L138). Presumably that should be fixed, though it isn't clear whether `glibc` has added support for `posix_spawn_file_actions_addchdir` (without `_np`) yet; perhaps we should test for both the portable and non-portable versions?

Since there may be other cases where `posix_spawn` cannot be used, I wonder if we should retain `vfork` support as well (preferring `posix_spawn` if possible, and falling back on `fork` if `vfork` is not available).

cc @bgamari

bgamari:
I'm rather reluctant to reintroduce `vfork` support for a few reasons:

1. In practice, it is quite tricky to use correctly since you need to guarantee that the forked child does not touch any memory of the forking parent before `exec`'ing. While in principle `runProcess` is a fairly simple piece of code which can be easily audited, it calls various bits of `libc` which are much harder to say anything about.
2. It is easy to end up with privilege escalation issues when `vfork` is used in a multithreaded program which uses `vfork` to spawn a child with reduced privileges.
3. POSIX.1-2008 has formally removed any specification of its behavior, making it a bit hard to depend upon

Ultimately, I wouldn't necessarily be opposed to reintroducing support assuming that we explicitly disable its use when `setuid` or `setgid` are requested. This would be sufficient to side-step (1) and (2), which should make it fairly safe (assuming its semantics under POSIX.1-2001).

adamgundry:
I don't claim to have a great amount of expertise here, so I'm happy to trust your judgement. I know there are cases in which `posix_spawn` isn't used so we have to fall back to `[v]fork`, but I don't know how common those are (at least once #303 resolves the issue with `posix_spawn_file_actions_addchdir`). If they are sufficiently rare, perhaps it is indeed fine to stay away from `vfork`.

[...]

PR #304, opened 2024-02-08
posix_spawn: Add support for posix_spawn_file_actions_addchdir_np

bgamari:
It appears that glibc still has not introduced support for `posix_spawn_file_actions_addchdir`, despite it being ratified by Austin Group. Add support for `posix_spawn_file_actions_addchdir_np` as a stop-gap solution since the removal of the `vfork` path has turned up some pathological cases in the `fork` path in user code.

Also fix a typo which rendered the previous `posix_spawn_file_actions_addchdir_np` path unusable.


adamgundry:
FWIW I raised #303 with a bit more background.

Issue #235, opened 2022-02-21
Clarify behaviour of `close_fds` on Windows


robx:
The documentation of `close_fds` states that:

> Close all file descriptors except stdin, stdout and stderr in the new process (on Windows, only works if std_in, std_out, and std_err are all Inherit).

It's quite unclear what "only works" means. Will things blow up if I set `close_fds` and `CreatePipe`? My best guess is something like "if any of `std_in`, `std_out`, `std_err` are not `Inherit`, those will be closed too".

Specifically, based on the documentation, I can't answer the question whether

```
(rstdin, wstdin) <- createPipe
(_,_,_,p) <- createProcess (proc "cat" []) { std_in = UseHandle rstdin, close_fds = True }
hPutStr wstdin "foo"
hClose wstdin
waitProcess p
```

should work on Windows. (This hangs on Linux without `close_fds`, (apparently) because the child inherits the open write-end, causing the pipe to stay open despite `hClose`.)

snoyberg:
@Mistuke  do you have any thoughts here?

robx:
Thank you for the detailed explanation. I see how that is hard to document clearly :/

I like the general idea of providing an API where `createPipe` generally has "non-inheritable" handles -- my understanding that could also work be done on (some?) unix systems by setting CLOEXEC.

(I've solved my particular problem that prompted this issue by reverting to use `CreatePipe` instead of `UseHandle` -- don't actually know in how far the code works right on Windows, but I'm no longer potentially breaking things.)

PR #308, opened 2024-04-02
Add CommunicationHandle, an API for inter-process communication via Handles

sheaf:
**NB** This PR depends on #310. For review, please only look at the last commit (at the time of writing, this was a6af24e973249f8992c274feae382a5d341ad02b).

This PR adds the `System.Process.CommunicationHandle` module, which provides the cross-platform `CommunicationHandle` abstraction which allows `Handle`s to be passed to child processes for inter-process communication.

A high-level API is provided by the function `readCreateProcessWithExitCodeCommunicationHandle`, which can be consulted for further details about how the functionality is meant to be used.

To test this functionality, I added the new `cli-child` executable component to the library. To work around [Cabal bug #9854](https://github.com/haskell/cabal/issues/9854), it was necessary to change the build-type of the library to `Custom`, in order to make the `cli-child` executable visible when running the test-suite. The custom `Setup.hs` script contains more details about the problem.

TODO:

  - [x] On Windows using WinIO, associating the parent handles to the parent process triggers `associateHandleWithIOCP: invalid argument (The parameter is incorrect.)`. (To reproduce, run `cabal run test -- +RTS -io-manager=native`; comment out all tests except `testCommunicationHandle` in `test/main.hs` for convenience.)
    - Fixed by properly creating a non-inheritable handle for the parent.
  - [x] On Windows, when NOT using WinIO, if we change the child from `let !output = force $ (reverse $ take 5 input) ++ "123"` to `let !output = force $ (reverse input) ++ "123"`, i.e. waiting for the end of the stream, then the test loops indefinitely.
    - Fixed by using named pipe always, instead of using anonymous pipes when not using WinIO.
  - [x] Get CI working: done, see #310.

### DIFF
--- a/System/Process/Common.hs
+++ b/System/Process/Common.hs
@@ -19,6 +19,7 @@ module System.Process.Common
     , mbFd
     , mbPipe
     , pfdToHandle
+    , rawFdToHandle
 
 -- Avoid a warning on Windows
 #if defined(mingw32_HOST_OS)
@@ -27,19 +28,22 @@ module System.Process.Common
     , CGid
 #endif
 
--- WINIO is only available on GHC 8.12 and up.
-#if defined(__IO_MANAGER_WINIO__)
+#if defined(mingw32_HOST_OS)
     , HANDLE
+-- WINIO is only available on GHC 9.0 and up.
+#  if defined(__IO_MANAGER_WINIO__)
     , mbHANDLE
     , mbPipeHANDLE
+    , rawHANDLEToHandle
+#  endif
 #endif
     ) where
 
 import Control.Concurrent
 import Control.Exception
-import Data.String
+import Data.String ( IsString(..) )
 import Foreign.Ptr
-import Foreign.Storable
+import Foreign.Storable ( Storable(peek) )
 
 import System.Posix.Internals
 import GHC.IO.Exception
@@ -278,8 +282,11 @@ mbPipe CreatePipe pfd  mode = fmap Just (pfdToHandle pfd mode)
 mbPipe _std      _pfd _mode = return Nothing
 
 pfdToHandle :: Ptr FD -> IOMode -> IO Handle
-pfdToHandle pfd mode = do
-  fd <- peek pfd
+pfdToHandle pfd mode =
+  ( \ fd -> rawFdToHandle fd mode ) =<< peek pfd
+
+rawFdToHandle :: FD -> IOMode -> IO Handle
+rawFdToHandle fd mode = do
   let filepath = "fd:" ++ show fd
   (fD,fd_type) <- FD.mkFD (fromIntegral fd) mode
                        (Just (Stream,0,0)) -- avoid calling fstat()
@@ -292,6 +299,11 @@ pfdToHandle pfd mode = do
   let enc = localeEncoding
 #endif
   mkHandleFromFD fD' fd_type filepath mode False {-is_socket-} (Just enc)
+
+
+#if defined(mingw32_HOST_OS) && !defined(__IO_MANAGER_WINIO__)
+type HANDLE = Ptr ()
+#endif
 
 #if defined(__IO_MANAGER_WINIO__)
 -- It is not completely safe to pass the values -1 and -2 as HANDLE as it's an
@@ -307,11 +319,14 @@ mbHANDLE _std NoStream        = return $ intPtrToPtr (-2)
 mbHANDLE _std (UseHandle hdl) = handleToHANDLE hdl
 
 mbPipeHANDLE :: StdStream -> Ptr HANDLE -> IOMode -> IO (Maybe Handle)
-mbPipeHANDLE CreatePipe pfd  mode =
-  do raw_handle <- peek pfd
-     let hwnd  = fromHANDLE raw_handle :: Io NativeHandle
-         ident = "hwnd:" ++ show raw_handle
-     enc <- fmap Just getLocaleEncoding
-     Just <$> mkHandleFromHANDLE hwnd Stream ident mode enc
+mbPipeHANDLE CreatePipe pfd mode =
+  Just <$> ( ( \ hANDLE -> rawHANDLEToHandle hANDLE mode ) =<< peek pfd )
 mbPipeHANDLE _std      _pfd _mode = return Nothing
+
+rawHANDLEToHandle :: HANDLE -> IOMode-> IO Handle
+rawHANDLEToHandle raw_handle mode  = do
+  let hwnd  = fromHANDLE raw_handle :: Io NativeHandle
+      ident = "hwnd:" ++ show raw_handle
+  enc <- getLocaleEncoding
+  mkHandleFromHANDLE hwnd Stream ident mode (Just enc)
 #endif
--- a/System/Process/Common.hs
+++ b/System/Process/Common.hs
@@ -21,7 +21,7 @@ module System.Process.Common
     , pfdToHandle
 
 -- Avoid a warning on Windows
-#ifdef WINDOWS
+#if defined(mingw32_HOST_OS)
     , CGid (..)
 #else
     , CGid
@@ -63,7 +63,7 @@ import GHC.JS.Prim (JSVal)
 
 -- We do a minimal amount of CPP here to provide uniform data types across
 -- Windows and POSIX.
-#ifdef WINDOWS
+#if defined(mingw32_HOST_OS)
 import Data.Word (Word32)
 import System.Win32.DebugApi (PHANDLE)
 #if defined(__IO_MANAGER_WINIO__)
@@ -75,7 +75,7 @@ import System.Posix.Types
 
 #if defined(javascript_HOST_ARCH)
 type PHANDLE = JSVal
-#elif defined(WINDOWS)
+#elif defined(mingw32_HOST_OS)
 -- Define some missing types for Windows compatibility. Note that these values
 -- will never actually be used, as the setuid/setgid system calls are not
 -- applicable on Windows. No value of this type will ever exist.
--- a/process.cabal
+++ b/process.cabal
@@ -1,14 +1,14 @@
+cabal-version: 2.4
 name:          process
 version:       1.6.19.0
 -- NOTE: Don't forget to update ./changelog.md
-license:       BSD3
+license:       BSD-3-Clause
 license-file:  LICENSE
 maintainer:    libraries@haskell.org
 bug-reports:   https://github.com/haskell/process/issues
 synopsis:      Process libraries
 category:      System
 build-type:    Configure
-cabal-version: >=1.10
 description:
     This package contains libraries for dealing with system processes.
     .
@@ -18,9 +18,11 @@ description:
     read more about it at
     <https://github.com/fpco/typed-process/#readme>.
 
+extra-doc-files:
+    changelog.md
+
 extra-source-files:
     aclocal.m4
-    changelog.md
     configure
     configure.ac
     include/HsProcessConfig.h.in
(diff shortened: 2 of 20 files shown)


sheaf:
@Mistuke I have mostly taken the implementation from the following two modules in the GHC source code:

- [`GHC.Runtime.Utils.runWithPipes`](https://gitlab.haskell.org/ghc/ghc/-/blob/master/compiler/GHC/Runtime/Utils.hs#L25) for the parent process,
- [`GHCi.Utils`](https://gitlab.haskell.org/ghc/ghc/-/blob/master/libraries/ghci/GHCi/Utils.hsc) for the child process.

I have attempted to debug the two TODOs described in the OP with @bgamari. In particular, we hope you'd be able to comment about the WinIO-specific aspects; in particular, it seems that creating a pipe (using the `process` `createPipe` function) and then immediately associating one end with the parent using `associateHandle'` seems to trigger `associateHandleWithIOCP: invalid argument (The parameter is incorrect.)`, whereas associating the other end after it has been inherited by the child seems to work OK. I took that code from [here](https://gitlab.haskell.org/ghc/ghc/-/blob/master/compiler/GHC/Runtime/Utils.hs#L60)... is that code path currently tested?

sheaf:
I fixed the first bug, which is the same bug as [GHC #24618](https://gitlab.haskell.org/ghc/ghc/-/issues/24618) (as I got it from copying the code from [`GHC.Runtime.Utils.runWithPipes`](https://gitlab.haskell.org/ghc/ghc/-/blob/master/compiler/GHC/Runtime/Utils.hs#L25)). Now we correctly only inherit one side of the pipe, and we don't call `associateHandle'` on the inherited handle.

The second bug is likely a bug in the compatibility `unix` code for Windows, but I haven't investigated.

bgamari:
@sheaf the Windows issues when using the old IO manager may indeed be attributable due to an infelicity in `msvcrt`'s POSIX IO implementation, which is known to be suspect, as the logic in this MR works on both POSIX platforms and under WinIO.

sheaf:
@Mistuke Would it be possible for you to review the new `System.Process.CommunicationHandle` module? As mentioned above, it's based on the code in [`GHC.Runtime.Utils.runWithPipes`](https://gitlab.haskell.org/ghc/ghc/-/blob/master/compiler/GHC/Runtime/Utils.hs#L25) (for the parent process) and [`GHCi.Utils`](https://gitlab.haskell.org/ghc/ghc/-/blob/master/libraries/ghci/GHCi/Utils.hsc) (for the child process). It's intended to replace that code, fixing [GHC bug #24618](https://gitlab.haskell.org/ghc/ghc/-/issues/24618) in the process.

process.cabal
```diff
@@ -1,14 +1,14 @@
+cabal-version: 2.4
 name:          process
-version:       1.6.18.0
```
bgamari, line 2: If I'm not mistaken this isn't *necessarily* a major version bump under the PVP since things have only been added. I'm rather of mixed feelings about this; it seems wrong for such a significant addition to only be a minor bump but on the other hand major bumps are quite a headache for everyone involved.

Bodigrim: It's common for core libraries to add lots of new functionality without bumping a major version, `filepath-1.4.100.0` is a prominent example of it. If there are no breaking changes, I'd suggest `1.6.19.0` or maybe `1.6.100.0` if it suits you better. Propagating major bumps through the ecosystem is way too expensive.

Mistuke (commented):
This looks on the whole good to me, note that this never made it into the docs but explains some of  the behavioral difference https://github.com/haskell/process/issues/235#issuecomment-1048057373

I do have one question, since you're making a new API, I wonder, should make the distinction between HANDLE and FDs?  I think it would be better to only use HANDLE.  i.e. only have `usingPipes`. Named pipes should be readable through the same interfaces as anonymous ones.  It's the inverse that's not true.

sheaf:
> So first thing to note, is that you cannot use async APIs (even in synchronous mode) with anonymous pipes.
So if the handle is created with _pipe there's no way for WinIO to handle it.

What I don't understand is that, with the first version of this PR, it worked fine for the parent to not use WinIO yet for the child to use WinIO. The parent will use `createPipeFd` (i.e. `_pipe` and not `mkNamedPipe`), call `_get_osfhandle` on that to get a `HANDLE`, pass that to the child, and on the other end (now using WinIO) we call `getGhcHandleNative :: HANDLE -> IO Handle` and we are able to read/write from the handle.

> I do have one question, since you're making a new API, I wonder, should make the distinction between HANDLE and FDs? I think it would be better to only use HANDLE. i.e. only have `usingPipes`. Named pipes should be readable through the same interfaces as anonymous ones. It's the inverse that's not true.

I have pushed a commit that uses only `mkNamedPipe` and not `_pipe`. See `createCommunicationPipe`. However I'm not sure how to use the `HANDLE` that `mkNamedPipe` creates to create a `Handle` on a GHC that doesn't have WinIO support (i.e. I need the `System.Process.Common.mbPipeHANDLE` function even on a GHC that doesn't support WinIO). Currently I am doing `open_osfhandle :: HANDLE -> IO Fd`,  `mkFd :: Fd -> ... -> IO FD`, `mkFileHandle :: FD -> ... -> IO Handle`, but I don't know if that is throwing the baby out with the bathwater.

sheaf:
Using `mkNamedPipe` has fixed the second bug reported in the OP, wherein the child process would hang waiting for the end of the stream when the parent was not using WinIO.

System/Process/CommunicationHandle.hsc
```diff
@@ -0,0 +1,375 @@
...
+  return () <!>
+    (associateHandle' ch `catch` handleAssociateHandleIOError ch)
+##endif
+  getGhcHandle ch
+
+##if defined(__IO_MANAGER_WINIO__)
+-- Internal function used when associating a 'HANDLE' with the current process.
+--
+-- Explanation: with WinIO, a synchronous handle cannot be associated with the
+-- current process, while an asynchronous one must be associated before being usable.
```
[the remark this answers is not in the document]
sheaf: I have tried re-opening the file but it never seems to work. I'm leaving that for a future PR: see #309.

process.cabal
sheaf: Thanks for your input; I reverted back to using a minor version bump.

System/Process/CommunicationHandle.hsc
```diff
@@ -0,0 +1,371 @@
...
+  -- The following implementation parallels 'readCreateProcess'
+  withCreateProcess cp $ \ _ _ _ ph -> do
+    -- Close the parent's references to the 'CommunicationHandle's after they
+    -- have been inherited by the child (we don't want to keep pipe ends open).
+    closeCommunicationHandle chTheyWrite
+    closeCommunicationHandle chTheyRead
+
+    -- Fork off a thread that waits on the output.
+    output <- readAction hWeRead
+    withForkWait (evaluate $ rnf output) $ \ waitOut -> do
```
tomjaguarpaw, line 364: `rnf` looks suspicious. What is it for?

Is it for forcing all lazy IO before closing `hWeRead`? If so then that's merely a sticking plaster, not a proper solution, because I can easily subvert the `NFData` instance by writing my own on a `newtype` wrapper. Therefore I would prefer not to impose the `NFData` constraint at all.

If it's for something else, could you add a comment to explain?

https://github.com/haskell/process/pull/308/files#diff-edde8bd0b4546b5a2c8aad3f7a290cec8d6c6cd698e7c5d707be3a2a847e446cR364

System/Process/Common.hs
```diff
@@ -270,8 +272,11 @@ mbPipe CreatePipe pfd  mode = fmap Just (pfdToHandle pfd mode)
 mbPipe _std      _pfd _mode = return Nothing
 
 pfdToHandle :: Ptr FD -> IOMode -> IO Handle
-pfdToHandle pfd mode = do
-  fd <- peek pfd
+pfdToHandle pfd mode =
+  rawFdToHandle mode =<< peek pfd
+
+rawFdToHandle :: IOMode -> FD -> IO Handle
+rawFdToHandle mode fd = do
```
tomjaguarpaw, line 279: It's a bit strange that `rawFdToHandle` has opposite argument order to `pfdToHandle`.

tomjaguarpaw, line 279: It's a bit strange that `rawFdToHandle` has opposite argument order to `pfdToHandle`.

tomjaguarpaw:
Thanks, this all looks fine from my point of view, which is basically the API level, since I don't have particular expertise in the internals.

System/Process/CommunicationHandle.hsc
sheaf: This logic is copied from [`readCreateProcess`](https://github.com/haskell/process/blob/master/System/Process.hs#L520), except that that function uses `String` to communicate with the child process. I think we should not decide for the user of the API which type we use for inter-process communication (and potential users such as `GHC` certainly do not want to just use `String`), so I made the code parametric over this type (requiring an `NFData` instance for that type).

I agree that all this use of lazy IO makes me a bit uneasy, but I preferred to re-use the existing implementation that exists in process rather than doing it completely differently. I am not against taking a different approach that avoids lazy IO, but perhaps that can be the topic of another MR that updates the existing code as well? (I am not necessarily opposed to only changing this code path, but being consistent across the library in this regard seems more sensible to me.)

tomjaguarpaw: Ah yes, I see. I agree it makes sense to have this new code be consistent with existing code.

In fact, can't we always use this API by setting `a ~ ()`? Instead of returning a complete value it seems like we can just act on the data we receive incrementally in `readAction`.

System/Process/CommunicationHandle.hsc
```diff
@@ -0,0 +1,371 @@
...
+-- > readCreateProcessWithExitCodeCommunicationHandle
+-- >   (\(chTheyRead, chTheyWrite) -> proc "child-exe" [show chTheyRead, show chTheyWrite])
+-- >   (\ hWeRead -> hGetContents hWeRead)
+-- >   (\ hWeWrite -> hPut hWeWrite "xyz")
+--
+-- where @child-exe@ is a separate executable that is implemented as:
+--
+-- > main = do
+-- >   [chRead, chWrite] <- getArgs
+-- >   hRead  <- useCommunicationHandle $ read chRead
```
andreabedini, line 334: `useCommunicationHandle` is not exported by the module.

tomjaguarpaw: Ah yes, mentions of `useCommunicationHandle` should probably be `openCommunicationHandleRead` and `openCommunicationHandleWrite`.

System/Process/CommunicationHandle.hsc
sheaf: I'm not sure about your suggestion to use `a ~ ()`. Perhaps consumers can modify their code if the API was changed in that way, but I preferred to keep the API the same as for `readCreateProcess`, just with different types than `String`. I hope that's OK.

sheaf:
I think this should be good to go now. I was just waiting for the lower bounds to the `time` package to be updated, as incorrect bounds were causing CI to fail. CI should be reliably good now.

System/Process/CommunicationHandle.hsc
tomjaguarpaw: Yes, that's fine.

My comment about `a ~ ()` was that I think that no value of `a` other than `()` is actually *required* to use this API. You can simulate other desired behavior for other values of `a` with a carefully chosen `readAction`. But I agree with the point that consistency with existing functions is important.

PR #310, opened 2024-04-05
Move tests to separate process-tests package & update CI to use GitHub actions

sheaf:
The first commit moves the test-suite to a separate package. The rationale is that other work, such as #308, requires the test-suite to use a Custom setup in order to work around [Cabal bug #9854](https://github.com/haskell/cabal/issues/9854). If we don't split up the test-suite into a separate package, this would introduce a setup-depends dependency on the Cabal library, which would create a gnarly dependency cycle between `Cabal` and `process`.

The second commit rewrites the CI script. I did this because I couldn't figure out how to provision the `Cabal` library (needed by #308) across GHC versions with the current CI infrastructure.

There are a few other auxiliary changes, such as:

  - Using Cabal version >= 2.4 in the .cabal files, and fixing the associated warnings,
  - Using the CPP mingw32_HOST_OS define to check for Windows;
    this avoids having to define the same variable twice in two different packages.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -3,47 +3,114 @@ name: Tests
 on:
     pull_request:
     push:
-        branches:
-        - master
+      branches:
+        - '**'
 
 jobs:
   build:
-    name: CI
+    name: GHC ${{ matrix.ghc-version }} on ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
     strategy:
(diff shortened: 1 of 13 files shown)


sheaf:
This PR is ready now. The previously occurring build failures were due incorrect lower bounds in the `time` package, which have since been addressed by a Hackage revision.

tomjaguarpaw (approved):

PR #316, opened 2024-05-11
Bump to base < 4.21 (to support GHC 9.10)

tomjaguarpaw:
I don't know what we should do about testing this, since `haskell-actions` doesn't have 9.10 yet. Should we wait for it to do so?

(Relatedly, I'm surprised that, since `process` is a boot library, that GHC could even release without this bump having already been performed. Maybe GHC just doesn't look at cabal version bounds at all.)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -20,6 +20,7 @@ jobs:
           - { os: windows-latest, arch: x64 }
         ghc-version:
           - 'latest'
+          - '9.10'
           - '9.8'
           - '9.6'
           - '9.4'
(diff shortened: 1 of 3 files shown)


tomjaguarpaw:
GHC 9.10 is now available in `haskell-actions` so I plan to merge this change to CI and make a Hackage revision.

tomjaguarpaw:
Revision 1 made: https://hackage.haskell.org/package/process-1.6.20.0/revisions/

bgamari:
@tomjaguarpaw, in the case of strict version bounds bumps it's not uncommon for GHC to carry patches locally performing the bounds bumps. This has historically been necessary since some core library authors are reluctant to merge bounds bumps until the GHC release has been made final.

tomjaguarpaw:
That makes sense. Thanks for the explanation!

Issue #317, opened 2024-06-20
`process-1.6.19.0`: relax upper bound of dependency on `base`


mpilgrem:
`process-1.6.19.0` is a boot package of GHC 9.10.1 which is tied to `base-4.20.0.0`. However, the Cabal file for `process` on Hackage specifies `base < 4.20`.

I have been tryng to build Stack with GHC 9.10.1. It compiles with `process-1.6.19.0` if the upper bound of that package's dependency on `base` is ignored.

tomjaguarpaw:
If you're getting it from Hackage anyway, would `process-1.6.20.0` not be good enough? It supports `base < 4.21`

https://hackage.haskell.org/package/process-1.6.20.0

mpilgrem:
@tomjaguarpaw, you can build with `process-1.6.20.0` but that is a different build plan than the one promised by a snapshot that specifies GHC 9.10.1 and (by implication) the boot packages that come with GHC 9.10.1.

tomjaguarpaw:
That's interesting. Does Stack check bounds for boot libraries that come with GHC? Cabal (seemingly) doesn't. See: https://github.com/haskell/process/pull/316#issue-2290996712

mpilgrem:
It is a little complicated. If you specify a GHC boot package as an extra-dep then you have to specify also any boot package dependencies as an extra-dep. If you specify a boot package as an extra-dep, Stack effectively gets its Cabal file from Hackage.

Normally, you don't have to specify a GHC boot package as an extra-dep. That is because, normally, a Stackage snapshot does not specify any GHC boot packages directly. They get 'included' only because specifying a GHC version brings with it all of that version's boot packages.

Recently, however the Stackage project has been issuing mis-specified (from my perspective) `nightly` snapshots that specify `Cabal` directly. When means that if you override the compiler in those snapshots (I was overrding GHC 9.8.2 with GHC 9.10.1), you also have to specify the intended `Cabal` version as an extra-dep - and that leads to other boot packages needing to be specified as extra-deps.

tomjaguarpaw:
Interesting. In any case, this should do what you want: https://hackage.haskell.org/package/process-1.6.19.0/revisions/

andreasabel:
Close issue?

tomjaguarpaw:
@mpilgrem, is this issue resolved?

mpilgrem:
@tomjaguarpaw, it is - many thanks!

PR #318, opened 2024-06-21
Bump base to < 4.21

tomjaguarpaw:
Specifically on v1.6.19.0

### DIFF
--- a/process.cabal
+++ b/process.cabal
@@ -86,7 +86,7 @@ library
 
     ghc-options: -Wall
 
-    build-depends: base      >= 4.10 && < 4.20,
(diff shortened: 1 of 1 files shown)


tomjaguarpaw:
Related to https://github.com/haskell/process/issues/317

tomjaguarpaw:
I'm not actually certain this is testing what I think it is, because the PR is targeting `master`, which already has v1.6.20.0's changes in it.

tomjaguarpaw:
Seems to work locally so I've pushed tag v1.6.19.0-r2 and made a Hackage revision.

Issue #252, opened 2022-06-22
Support `AbstractFilePath`


hasufell:
## Motivation

The next filepath release will add support for a new API, which fixes subtle encoding issues and improves cross platform code and memory residence.

You can read about the state of the newly added API here: https://github.com/haskellfoundation/tech-proposals/issues/35

Release candidate with haddock is here: https://hackage.haskell.org/package/filepath-2.0.0.3/candidate

Demonstration about unsoundness of base with exotic encodings and how the new API fixes them is here: https://gist.github.com/hasufell/c600d318bdbe010a7841cc351c835f92

## Migrations

Migrations can happen once the `filepath`, `unix` and `Win32` packages are updated and in sync. A migration would usually involve using the new types from [System.AbstractFilePath](https://hackage.haskell.org/package/filepath-2.0.0.3/candidate/docs/System-AbstractFilePath.html) and the new API variants from [unix](https://github.com/haskell/unix/pull/202)/[Win32](https://github.com/haskell/win32/pull/198).

When writing low-level cross platform code manually (shouldn't generally be necessary), the usual strategy is this:

```hs
#if defined(mingw32_HOST_OS) || defined(__MINGW32__)
import qualified System.AbstractFilePath.Data.ByteString.Short.Word16 as SBS
import qualified System.AbstractFilePath.Windows as PFP
#else
import qualified System.AbstractFilePath.Data.ByteString.Short as SBS
import qualified System.AbstractFilePath.Posix as PFP
#endif

crossPlatformFunction :: AbstractFilePath -> IO ()
#if defined(mingw32_HOST_OS) || defined(__MINGW32__)
crossPlatformFunction (OsString pfp@(WS ba)) = do
    -- use filepath functions for windows specific operating system strings
    let ext = PFP.takeExtension pfp
    -- operate directly on the underlying bytestring (which is a wide character bytestring, so uses Word16)
    let foo = SBS.takeWhile
    ...
#else
crossPlatformFunction (OsString pfp@(PS ba)) = do
    -- use filepath functions for posix specific operating system strings
    let ext = PFP.takeExtension pfp
    -- operate directly on the underlying bytestring (which is just Word8 bytestring)
    let foo = SBS.takeWhile
    ...
#endif
```

[...]

PR #320, opened 2024-06-22
Posix: Do not treat arguments as `FilePath`

### LINKED ISSUE #319: NUL byte disallowed in arguments
The following program worked just fine when compiled with GHC <9.8:
```haskell
module Main (main) where

import System.Process

main :: IO ()
main = callProcess "echo" ["foo\0bar"]
```
Compiled with GHC >=9.8 it outputs the following runtime error:
```
*** Exception: /usr/bin/echo: checkForInteriorNuls: invalid argument (FilePaths must not contain internal NUL code units.)
```

The reason is that `process` also treats the arguments as `FilePath`: https://github.com/haskell/process/blob/d74bba244d9d7c83b3a0ef5159b4e2e0fa7df07b/System/Process/Posix.hs#L139 (Linking to the POSIX implementation here since I am running Linux.)
The `System.Posix.Internals.withFilePath` function referenced there checks for internal NULL characters since `base >4.19` due to the accepted CLC propsal https://github.com/haskell/core-libraries-committee/issues/144.

mmhat:
Fixes #319

### DIFF
--- a/System/Process/Posix.hs
+++ b/System/Process/Posix.hs
@@ -136,7 +136,9 @@ createProcess_Internal fun
    maybeWith withFilePath mb_cwd $ \pWorkDir ->
    maybeWith with mb_child_group $ \pChildGroup ->
    maybeWith with mb_child_user $ \pChildUser ->
-   withMany withFilePath (cmd:args) $ \cstrs ->
+   withFilePath cmd $ \cmdstr ->
+   withMany withCString args $ \argstrs -> do
+   let cstrs = cmdstr : argstrs
    withArray0 nullPtr cstrs $ \pargs -> do
 
      fdin  <- mbFd fun fd_stdin  mb_stdin


tomjaguarpaw:
This looks pretty plausible to me, but I'm not an expert with the low level details. There are various people who are better qualified. Perhaps one of them can chime in.

@bgamari @hasufell @snoyberg @Mistuke (or someone else?)

hasufell:
There are two differences:

1. `withFilePath` uses `getFileSystemEncoding`, `withCString` uses `getForeignEncoding`
2. `withFilePath` checks for interior nulls now (not sure since which base version) and barfs out in case it finds any

Tbh, all of those s*ck. But that's another discussion: https://github.com/haskell/process/issues/252

System/Process/Posix.hs
```diff
@@ -136,7 +136,9 @@ createProcess_Internal fun
    maybeWith withFilePath mb_cwd $ \pWorkDir ->
    maybeWith with mb_child_group $ \pChildGroup ->
    maybeWith with mb_child_user $ \pChildUser ->
-   withMany withFilePath (cmd:args) $ \cstrs ->
+   withFilePath cmd $ \cmdstr ->
+   withMany withCString args $ \argstrs -> do
```
Mistuke, line 140: On Windows this would be wrong, it needs to be `withCWString` otherwise it's encoded as 8-byte string rather than 16-bytes.

mmhat: @Mistuke Thanks for the feedback; Then `withCWString` is indeed the correct choice as I suspected here: https://github.com/haskell/process/issues/319#issuecomment-2184005410

mmhat:
> 2. `withFilePath` checks for interior nulls now (not sure since which base version) and barfs out in case it finds any

I think it is since `base-4.18.0.0` (GHC 9.6.1).

If there are no further objections: Who is willing to merge this?

System/Process/Posix.hs
tomjaguarpaw: What is the conclusion of this exchange? On Windows it would be wrong, but it's not Windows, right? It's POSIX. So is `withCString` indeed right in this case?

hasufell: > What is the conclusion of this exchange? On Windows it would be wrong, but it's not Windows, right? It's POSIX. So is `withCString` indeed right in this case?

Windows is already correct:

https://github.com/haskell/process/blob/302b43a2e0c514a6ad86cbc402c60bf7f86878e2/System/Process/Windows.hsc#L185

This PR can be merged.

tomjaguarpaw:
OK, released as https://hackage.haskell.org/package/process-1.6.22.0. Thanks @mmhat!

PR #321, opened 2024-06-25
Remove `includes` from `.cabal`-file

sol:
This is useless at best, and a common source of confusion.

https://github.com/haskell/cabal/pull/10145
https://github.com/sol/hpack/issues/355

### DIFF
--- a/process.cabal
+++ b/process.cabal
@@ -82,8 +82,6 @@ library
             build-depends: unix >= 2.5 && < 2.9
 
     include-dirs: include
-    includes:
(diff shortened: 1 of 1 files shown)


tomjaguarpaw:
This has been there for a very long time! It first appeared in d7dbde724dc5654fe4007cef1ba97a1bf3bb7714. Reading the links you provided it appears that `includes` is completely ignored in the modern world. I'm in favour of merging your MR and will do so in due course if no one else objects.

PR #322, opened 2024-07-25
Don't attach finalizers to Handles in CommunicationHandle API

sheaf:
We are now careful to not attach any finalizers to Handles when creating pipes for inter-process communication on Unix systems. Instead, the handles are closed manually.

The finalizers were causing problems in situations such as the following:

  - the parent creates a new pipe, e.g. pipe2([7,8]),
  - the parent spawns a child process, and lets FD 8 be inherited by the child,
  - the parent closes FD 8 (as it should),
  - the parent opens FD 8 for some other purpose, e.g. for writing to a file,
  - the finalizer for the Handle wrapping FD 8 runs, closing FD 8, even though it is now in use for a completely different purpose.

This PR does not include a test, as the above bug is a bit difficult to trigger. My reproducer for this bug was a test in the `cabal-install` testsuite; I can confirm that the bug no longer occurs with this patch.

This PR bumps the process version to `1.6.21.0`. After releasing it on Hackage, I would also recommend deprecating `1.6.20.0`.

### DIFF
--- a/System/Process/CommunicationHandle.hs
+++ b/System/Process/CommunicationHandle.hs
@@ -38,12 +38,18 @@ import Control.DeepSeq (NFData, rnf)
 -- | Turn the 'CommunicationHandle' into a 'Handle' that can be read from
 -- in the current process.
 --
+-- The returned 'Handle' does not have any finalizers attached to it;
+-- use 'hClose' to close it.
+--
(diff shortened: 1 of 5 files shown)


tomjaguarpaw:
This seems sensible!

PR #324, opened 2024-09-04
Prevent command-line injection for batch files with trailing char

TristanCacqueray:
This change ensure the regime implemented for HSEC-2024-0003 is applied to batch file names ending with trailing chars that are ignored by Windows.

### DIFF
--- a/System/Process/Windows.hsc
+++ b/System/Process/Windows.hsc
@@ -28,6 +28,7 @@ import Control.Exception
 import Control.Monad
 import Data.Bits
 import Data.Char (toLower)
+import Data.List (dropWhileEnd)
 import Foreign.C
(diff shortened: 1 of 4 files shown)


TristanCacqueray:
Note that the recommended fix is to use `GetFullPathNameW` to get the real file name. Unfortunately I don't have a Windows system to implement this.

I believe the fix proposed here works, but perhaps it can produce false positive if a legitimate non batch script ends with `.bat ` or something. Though I don't know if that's even possible since the trailing chars are also ignored when saving a file.

bgamari:
Thank you for handling this, @TristanCacqueray !

TristanCacqueray:
You are welcome!

Issue #327, opened 2024-09-29
Fix javascript build


bgamari:
Currently `master` is not buildable with the Javascript backend due to
```
libraries/process/System/Process/CommunicationHandle/Internal.hsc:26:8: error: [GHC-87110]
    Could not load module ‘System.Posix’.
    It is a member of the hidden package ‘unix-2.8.4.0’.
    Use -v to see a list of the files searched for.
   |
26 | import Control.Exception (catch, throwIO)
   |        ^^^^^^^^^^^^
```

bgamari:
In particular, the problematic imports are
```haskell
import System.Posix
  ( Fd(..)
  , FdOption(..), setFdOption
  )
import System.Posix.Internals
  ( fdGetMode )
```

bgamari:
It looks like `setFdOption` is available with the Javascript backend. Consequently I believe simply adding a `unix` dependency should resolve this.

PR #329, opened 2024-09-30
 Fix build with Javascript backend 

bgamari:
The new `CommunicationsHandle` abstraction requires the `setFdOptions` from `unix`.

Fixes https://github.com/haskell/process/issues/327.


bgamari:
I have confirmed that this fixes the GHC/Javascript build; I will release shortly.

hsyl20:
Thanks for fixing this. I was about to report this bug :)

PR #330, opened 2024-11-22
Fix haddocks of `createPipe`

sol:


System/Process/Internals.hs
```diff
@@ -200,22 +200,22 @@ runGenProcess_ fun c _ _ = createProcess_ fun c
...
 --     #endif
 --
 --     ...
 --
---     #if defined (__IO_MANAGER_WINIO__)
---     return () <!> (do
---       associateHandle' =<< handleToHANDLE <handle>)
+--     #if defined(__IO_MANAGER_WINIO__)
+--     return () `<!>` do
+--       associateHandle' `=<<` handleToHANDLE readEnd
```
tomjaguarpaw, line 212: Would you mind posting a before and after screenshot? I'm confused what the backticks are for (escaping the `<`?).

sol: > I'm confused what the backticks are for (escaping the `<`?).

They hyperlink the identifier.   The main reason I used them here is that otherwise `<` would need to be escaped as `\<`.   That they also hyperlink  is a nice by product.

> Would you mind posting a before and after screenshot?

Sorry, I'm away from my computer right now.

Please look at Hackage for how it is currently rendered.  You will immediately see that it is broken. This PR fixes this. You can verify this with `cabal haddock`.

tomjaguarpaw: OK, sure. I don't think we need to hyperlink `=<<`, and the link didn't appear on `<!>` anyway, so I converted it to escaping.

### Before

screenshot

### After

screenshot2

tomjaguarpaw:
Thanks!

PR #332, opened 2025-01-01
Fix support for GHCJS_BROWSER compilation flag of GHC

Swordlash:
See https://gitlab.haskell.org/ghc/ghc/-/issues/25613#note_602248

When compiling `ghc` under GHCJS_BROWSER, all the environment checking code should be removed by the preprocessor. This is not the case for the `jsbits` of the `process` library, which causes troubles for Google Closure Compiler and bundlers. This PR adds missing directives.


Swordlash:
I'm going to close this PR for now since we agreed to remove the macro altogether.

PR #336, opened 2025-02-16
Remove extraneous Haddocks comment

thomasjm:
This was added in #292. CC @luite


tomjaguarpaw:
Thanks. It would be good to keep the `XXX` text as a comment.

tomjaguarpaw:
Lovely, thanks!

PR #337, opened 2025-02-18
for better memory efficiency

qxrein:
this change improves performance for large buffer writes and eliminates unnecessary memory allocation and copying
This creates a view into the existing buffer instead of copying data, which is more efficient for large writes.


bgamari:
Pinging @luite.

qxrein:
i am sorry for you

luite:
sorry, missed this

qxrein:
i even forgot what i did here

PR #339, opened 2025-03-30
Adds `getArgs :: IO [OsString]`

### LINKED ISSUE #338: Add `getArgs :: IO [OsString]`
With the [`Abstract FilePath Proposal`](https://github.com/haskellfoundation/tech-proposals/issues/35), we can now provide functions that work on the more appropriate `OsPath/OsString`, rather than legacy `type FilePath = [String]`. In particular, we really ought to have a version for [`System.Environment.getArgs :: IO [String]`](https://hackage.haskell.org/package/base-4.21.0.0/docs/System-Environment.html).

For example, see this [optparse-applicative issue](https://github.com/pcapriotti/optparse-applicative/issues/491).

Unfortunately, `OsString` cannot reside in `base`, hence this function has to live somewhere else. As this function is related to the current process, how about in `process`?

Luckily implementation should be simple, since the Win32/Posix functions already exist, so all that remains is an `OsString` shim.

- [`getArgs :: IO [PosixString]`](https://hackage.haskell.org/package/unix-2.8.6.0/docs/System-Posix-Env-PosixString.html#v:getArgs)
- [`getArgs :: IO [WindowsString]`](https://hackage.haskell.org/package/Win32-2.14.1.0/docs/System-Win32-WindowsString-Console.html#v:getArgs)

This issue is **not** for full [`OsString` compatibility](https://github.com/haskell/process/issues/252), though it is related. See also [this discussion](https://github.com/haskell/file-io/issues/41).

Thanks!

CC @hasufell

tbidne:
Resolves #338.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,12 @@
 # Changelog for [`process` package](http://hackage.haskell.org/package/process)
 
+## X.Y.Z
+
+* Add `System.Process.Environment.OsString`.
+* Bumps `base >= 4.12.0.0` (GHC 8.6+), `filepath >= 1.4.100.0`,
+  `unix >= 2.8.0.0`, and `Win32 >= 2.14.1.0`.
+* Drops support for GHC < 8.6.
+
 ## 1.6.25.0 *September 2024*
 
 * Fix build with Javascript backend ([#327](https://github.com/haskell/process/issues/327))
--- a/process.cabal
+++ b/process.cabal
@@ -41,6 +41,11 @@ source-repository head
     type:     git
     location: https://github.com/haskell/process.git
 
+flag os-string
+    description: Use the new os-string package
+    default: False
+    manual: False
+
 library
     default-language: Haskell2010
     other-extensions:
@@ -56,19 +61,20 @@ library
         System.Process
         System.Process.CommunicationHandle
         System.Process.CommunicationHandle.Internal
+        System.Process.Environment.OsString
         System.Process.Internals
     other-modules: System.Process.Common
     if os(windows)
         c-sources:
             cbits/win32/runProcess.c
         other-modules: System.Process.Windows
-        build-depends: Win32 >=2.4 && < 2.15
+        build-depends: Win32 >= 2.14.1.0 && < 2.15
         -- ole32 and rpcrt4 are needed to create GUIDs for unique named pipes
         -- for process.
         extra-libraries: kernel32, ole32, rpcrt4
         cpp-options: -DWINDOWS
     else
-        build-depends: unix >= 2.5 && < 2.9
+        build-depends: unix >= 2.8.0.0 && < 2.9
         if arch(javascript)
             js-sources:
                 jsbits/process.js
@@ -88,7 +94,12 @@ library
 
     ghc-options: -Wall
 
-    build-depends: base      >= 4.10 && < 4.22,
+    build-depends: base      >= 4.12.0.0 && < 4.22,
                    directory >= 1.1 && < 1.4,
-                   filepath  >= 1.2 && < 1.6,
                    deepseq   >= 1.1 && < 1.6
+
+    if flag(os-string)
+        build-depends: filepath >= 1.5.0.0 && < 1.6,
+                       os-string >= 2.0.0 && < 2.1
+    else
+        build-depends: filepath >= 1.4.100.0 && < 1.5.0.0
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -29,15 +29,8 @@ jobs:
           - '8.10'
           - '8.8'
           - '8.6'
-          - '8.4'
-          - '8.2'
 
         exclude:
-          # Exclude GHC 8.2 on Windows (GHC bug: undefined reference to `__stdio_common_vswprintf_s')
-          - platform:
-              os: windows-latest
-            ghc-version: '8.2'
-
           # Only allow ARM jobs with GHC >= 9.2
           # (It's tedious to not be able to use matrix.ghc-version >= 9.2 as a conditional here.)
           - platform:
@@ -52,12 +45,6 @@ jobs:
           - platform:
               arch: arm
             ghc-version: '8.6'
-          - platform:
-              arch: arm
-            ghc-version: '8.4'
-          - platform:
-              arch: arm
-            ghc-version: '8.2'
 
     steps:
       - uses: actions/checkout@v4
--- a/System/Process/Environment/OsString.hs
+++ b/System/Process/Environment/OsString.hs
@@ -1,0 +1,41 @@
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE PackageImports #-}
+
+-- | Miscellaneous information about the system environment, for 'OsString'.
+--
+-- @since X.Y.Z
+module System.Process.Environment.OsString (
+    getArgs,
+    getEnv,
+    getEnvironment,
+    ) where
+
+import Data.Coerce (coerce)
+#if MIN_VERSION_filepath(1, 5, 0)
+import "os-string" System.OsString.Internal.Types (OsString(OsString))
+#else
+import "filepath" System.OsString.Internal.Types (OsString(OsString))
+#endif
+#if defined(mingw32_HOST_OS)
+import qualified System.Win32.WindowsString.Console as Platform
+#else
+import qualified System.Posix.Env.PosixString as Platform
+#endif
+
+-- | 'System.Environment.getArgs' for 'OsString'.
+--
+-- @since X.Y.Z
+getArgs :: IO [OsString]
+getArgs = coerce Platform.getArgs
+
+-- | 'System.Environment.getEnv' for 'OsString'.
+--
+-- @since X.Y.Z
+getEnv :: OsString -> IO (Maybe OsString)
+getEnv = coerce Platform.getEnv
+
+-- | 'System.Environment.getEnvironment' for 'OsString'.
+--
+-- @since X.Y.Z
+getEnvironment :: IO [(OsString, OsString)]
+getEnvironment = coerce Platform.getEnvironment


tomjaguarpaw:
This seems all fine to me, but I don't really "own" this library and don't know who does, so I don't know who gives final sign-off.

System/Process/Env/OsString.hs
```diff
@@ -0,0 +1,45 @@
...
+--
+-- Maintainer  :  libraries@haskell.org
+-- Stability   :  experimental
+-- Portability :  non-portable (requires concurrency)
+--
+-- Miscellaneous information about the system environment.
+--
+-- @since X.Y.Z
+--
+-----------------------------------------------------------------------------
```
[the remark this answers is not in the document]
tbidne: Indeed, the copied bit about portability (`non-portable (requires concurrency)`) also seems dubious if the comment is meant to apply to this module only. Happy to receive guidance here.

Bodigrim:
@tomjaguarpaw I'm pretty sure that you have all necessary authority to make decisions here, since Michael passed the torch. Otherwise you can reach to CLC for advice, but this PR does not look particularly controversial to me.

System/Process/Env/OsString.hs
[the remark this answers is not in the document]
Bodigrim: I'd skip all metadata fields here - they do not seem to add anything helpful.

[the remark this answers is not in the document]
tomjaguarpaw: I second that.

tbidne:
Two questions I have:

1. Do we need to do anything about js or wasm? I see `javascript_HOST_ARCH` and `wasm32_HOST_ARCH` pragmas used in various places. I tried the legacy string `getArgs` and unix `getArgs` on the js backend, and both give empty args without an error, so maybe it's fine?

2. What of the rest of `System.Environment`? I don't mean to feature-creep, but I think it would be helpful to have an idea of what is to become of the rest of that module. Does the argument in favor of adding `getArgs` here ("args pertaining to the current process") also apply to the environment and/or program name/path? If so, everything in that module could eventually go here.

    Right now, `getEnv :: String -> IO String` and `getEnvironment :: IO [(String, String)]` exist in `unix` and `Win32` for `PosixString` and `WindowsString`, respectively, so these could also be easily shimmed here in this PR. The others do not exist yet.

System/Process/Env/OsString.hs
```diff
@@ -0,0 +1,32 @@
+{-# LANGUAGE CPP #-}
+
+#if !MIN_VERSION_filepath(1,5,0)
+-- Silence the warning on System.OsString.Internal.Types import. It exists
+-- (deprecated) in filepath 1.4.100.0 && < 1.5.0.0. For filepath >= 1.5,
```
tomjaguarpaw, line 5:
```suggestion
-- (deprecated) in filepath >= 1.4.100.0 && < 1.5.0.0. For filepath >= 1.5,
```

Is there a missing `>=` here?

tomjaguarpaw, line 5: I don't see any deprecation notices here:

* https://hackage.haskell.org/package/filepath-1.4.100.0/docs/System-OsString-Internal-Types.html
* https://hackage.haskell.org/package/filepath-1.4.100.1/docs/System-OsString-Internal-Types.html
* https://hackage.haskell.org/package/filepath-1.4.100.2/docs/System-OsString-Internal-Types.html
* https://hackage.haskell.org/package/filepath-1.4.100.3/docs/System-OsString-Internal-Types.html
* https://hackage.haskell.org/package/filepath-1.4.100.4/docs/System-OsString-Internal-Types.html

and compiling with `cabal repl -b filepath==1.4.100.0` and this CPP removed does not emit a warning.

And in any case, I'm personally not fan of using CPP to suppress warnings. Do we really need this CPP?

process.cabal
```diff
@@ -90,5 +97,10 @@ library
 
     build-depends: base      >= 4.10 && < 4.22,
                    directory >= 1.1 && < 1.4,
-                   filepath  >= 1.2 && < 1.6,
                    deepseq   >= 1.1 && < 1.6
+
+    if flag(os-string)
+        build-depends: filepath >= 1.5.0.0 && <1.6,
+                       os-string >= 2.0.0 && <2.1
```
tomjaguarpaw, line 104: I don't really get what's going on here. This seems bad:

```
cabal repl -b os-string==2.0.7
___________________________________________________________________________________________________________Sat 5 Apr 13:08:41 BST 2025
Configuration is affected by the following files:
- cabal.project
Warning: The package list for 'hackage.haskell.org' is 34 days old.
Run 'cabal update' to get the latest list of available packages.
Resolving dependencies...
Build profile: -w ghc-9.6.6 -O1
In order, the following will be built (use -v for more details):
 - process-1.6.25.0 (interactive) (first run)
Preprocessing library for process-1.6.25.0...
GHCi, version 9.6.6: https://www.haskell.org/ghc/  :? for help
[1 of 9] Compiling System.Process.Common ( System/Process/Common.hs, interpreted )
[2 of 9] Compiling System.Process.Env ( System/Process/Env.hs, interpreted )
[3 of 9] Compiling System.Process.Env.OsString ( System/Process/Env/OsString.hs, interpreted )

System/Process/Env/OsString.hs:23:1: error:
    Ambiguous module name ‘System.OsString.Internal.Types’:
      it was found in multiple packages:
      filepath-1.4.300.1 os-string-2.0.7
   |
23 | import System.OsString.Internal.Types
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
[4 of 9] Compiling System.Process.Posix ( System/Process/Posix.hs, interpreted )
[5 of 9] Compiling System.Process.Internals ( System/Process/Internals.hs, interpreted )
[6 of 9] Compiling System.Process.CommunicationHandle.Internal ( dist-newstyle/build/x86_64-linux/ghc-9.6.6/process-1.6.25.0/build/System/Process/CommunicationHandle/Internal.hs, interpreted )
[7 of 9] Compiling System.Process   ( System/Process.hs, interpreted )
[8 of 9] Compiling System.Process.CommunicationHandle ( System/Process/CommunicationHandle.hs, interpreted )
[9 of 9] Compiling System.Cmd       ( System/Cmd.hs, interpreted )
Failed, 8 modules loaded.
```

System/Process/Env/OsString.hs
```diff
@@ -0,0 +1,32 @@
...
+    ) where
+
+import Data.Coerce (coerce)
+#if defined(mingw32_HOST_OS)
+import qualified System.Win32.WindowsString.Console as Platform
+import System.OsString.Internal.Types
+  ( OsString(OsString), WindowsString(WindowsString) )
+#else
+import System.OsString.Internal.Types
+  ( OsString(OsString), PosixString(PosixString) )
```
tomjaguarpaw, line 24: I don't understand why `WindowsString(WindowsString)` and `PosixString(PosixString)` are imported. Are they used? I don't see where. One might think they're needed for the `coerce` to work, but I think that only needs the `OsString` constructor. However, they don't emit redundant import warnings either, so I'm confused. In fact, if I change the import to

```.hs
import System.OsString.Internal.Types
 (OsString (OsString), PosixString)
```

then I *do* get a redundant import warning. Is GHC's redundant import checker just wrong when `coerce` is in play?

(I guess ideally we'd use [`coercionToPlatformTypes`](https://hackage.haskell.org/package/os-string-2.0.7/docs/System-OsString-Internal-Types.html#v:coercionToPlatformTypes). We shouldn't have to be diving into `os-string` internals here. But maybe we can't use it because we want to work with older `filepath`s?)

tomjaguarpaw:
> What of the rest of `System.Environment`?

I think if we're going to include one of its elements we may as well include them all.

System/Process/Env/OsString.hs
tbidne: Ah, it's in the 300+ series e.g. https://hackage.haskell.org/package/filepath-1.4.301.0/docs/System-OsString-Internal-Types.html.

In any case, no, it's not necessary. Removed.

process.cabal
tomjaguarpaw: This seems to fix things:

```diff
@ process.cabal:106 @ library
         build-depends: filepath >= 1.5.0.0 && <1.6,
                        os-string >= 2.0.0 && <2.1
     else
-        build-depends: filepath >= 1.4.100.0 && < 1.5.0.0
+        build-depends: filepath >= 1.4.100.0 && < 1.5.0.0,
+                       os-string < 0
```

```
cabal repl -b os-string==2.0.7
...
<succeeds>
```

That is to say, if you chose the "os-string: False" branch then you're not allowed to also use `os-string`-the-package.

System/Process/Env/OsString.hs
tomjaguarpaw: This doesn't seem to have been resolved. Has a push gone astray?

tomjaguarpaw: This doesn't seem to have been removed. Has a push gone astray?

process.cabal
tomjaguarpaw: (I suppose another workaround would be package imports.)

System/Process/Env/OsString.hs
tbidne: You are right. I reflexively imported `PosixString/WindowsString` as I thought it was needed for the coercion, but we only need `OsString`'s constructor.

The `OsString` constructor only exists in internal modules, so I'm not sure what else we can do. The coercion function you linked is interesting, but I think here cpp is preferable to runtime pattern matching.

tomjaguarpaw: The coercion function probably gets inlined and eliminated (if it doesn't that's a bug and we do whatever is required to fix it), but it's not available in older versions of our dependencies, so we'll have to do without it here.

System/Process/Env/OsString.hs
```diff
@@ -0,0 +1,36 @@
...
+--
+-- @since X.Y.Z
+getArgs :: IO [OsString]
+getArgs = coerce Platform.getArgs
+
+-- | 'System.Environment.getEnv' for 'OsString'.
+--
+-- @since X.Y.Z
+getEnv :: OsString -> IO (Maybe OsString)
+getEnv = coerce . Platform.getEnv . coerce
```
tomjaguarpaw, line 30:
```suggestion
getEnv :: OsString -> IO (Maybe OsString)
getEnv = coerce Platform.getEnv
```

The point of `coerce` is that you just `coerce`!

process.cabal
tbidne: `os-string < 0` seems like a good idea. @hasufell?

tbidne:
> > What of the rest of System.Environment?
>
> I think if we're going to include one of its elements we may as well include them all.

Okay, I've added the remaining functions (`getEnv` and `getEnvironment`) that currently exist in all 3 (`System.Environment`, `Win32`, and `unix`). Here is a list of outstanding functions that should probably also be here, once they exist in `Win32` and `unix`:

- `Env.getProgName :: IO OsString`
- `Env.executablePath :: Maybe (IO (Maybe OsString))`
- `Env.getExecutablePath :: IO OsString`
- `Env.lookupEnv :: OsString -> IO (Maybe OsString)`
- `Env.setEnv :: OsString -> OsString -> IO ()`
- `Env.unsetEnv :: OsString -> IO ()`
- `Env.withArgs :: [OsString] -> IO a -> IO a`
- `Env.withProgName :: String -> IO a -> IO a`

tbidne:
Some bikeshedding:

1. `System.Process.Env` (current) or `System.Process.Environment` (like `System.Environment`)?
2. Generic haddocks e.g. `'System.Environment.getArgs' for 'OsString'` or copied from `System.Environment`?

System/Process/Env/OsString.hs
tbidne: Possibly? I removed the entire comment per [your comment](https://github.com/haskell/process/pull/339#discussion_r2034063116). In any case, it's gone now.

process.cabal
hasufell: I'm actually not quite sure what the implication of this constraint is. But we should do it consistently for all boot packages.

E.g. there have been odd interactions with `--allow-newer`, which this is likely prone to as well.

`-b` in cabal is kind of a hack. You can break stuff arbitrarily, if you add any package via `-b` that exports the same modules as another package.

I actually think package imports might be a better option... e.g.

```hs
#if MIN_VERSION_filepath(1, 5, 0)
import "os-string" System.OsString.Internal.Types (OsString(OsString))
#else
import "filepath" System.OsString.Internal.Types (OsString(OsString))
#endif
```

hasufell: @Bodigrim

tomjaguarpaw:
1. I'd prefer keeping it as similar as possible to existing module names, so unless there's also something called `Env` that we're copying, I'd stick to `Environment`

process.cabal
Bodigrim: I'd prefer `PackageImports`, it's unambiguous.

tbidne: So do we keep the cabal flag too? We still need to ensure `os-string` is in `build-depends` when `filepath >= 1.5`.

I guess the alternative is to unconditionally require `os-string`. It will be superfluous in the `filepath < 1.5` case, but maybe that's better than mucking about with flags.

hasufell: We keep the flag and the constraints the same way other packages do (as in: without the `< os-string-0`).

tbidne:
Okay, renamed `Env -> Environment` and added the `MIN_VERSION_filepath(1, 5, 0)` cpp. Two CI questions:

1. Do we care about the failing `GHC 8.4 + Windows` job, due to [this](https://github.com/haskell/win32/pull/229)? To recap, the problem is that we now require a recent version of `Win32`, which unfortunately added `NumericUnderscores`, causing `GHC < 8.6.1` to fail.

    Ideally that PR would be merged and released, so CI here would go back to green once the solver starts picking the new release. But it's been open for a year, so that might not be anytime soon. The combination could be excluded here, if we want.

2. Do we want to test the flag on CI in some capacity? It's easy enough to add a new matrix for `os-string/no os-string`, though of course that basically doubles the number of jobs, from 41 to 82 (it's not _quite_ that bad, since some older combinations should be excluded). This is well within the free tier limit (256 I think), but still, it will make CI take longer.

process.cabal
tomjaguarpaw: Yeah, I think that's the right decision. `PackageImports` actually solves the particular problem, whereas `< 0` is a cute hack but only solves the problem accidentally.

changelog.md
```diff
@@ -1,5 +1,10 @@
 # Changelog for [`process` package](http://hackage.haskell.org/package/process)
 
+## X.Y.Z
+
+* Add `System.Process.Env` and `System.Process.Env.OsString`. Bumps
+  `filepath >= 1.4.100.0`.
```
tomjaguarpaw, line 6: This is now outdated. `s/Env/Environment/`?

process.cabal
```diff
@@ -56,19 +61,21 @@ library
         System.Process
         System.Process.CommunicationHandle
         System.Process.CommunicationHandle.Internal
+        System.Process.Environment
+        System.Process.Environment.OsString
```
tomjaguarpaw, line 65: I don't understand why we have two modules which export the exact same thing. What's that about?

tomjaguarpaw:
I suspect it's probably fine to give up on GHC 8.4.

process.cabal
hasufell: They don't. One uses `OsString`, one doesn't.

See:

* https://hackage.haskell.org/package/filepath-1.5.4.0/docs/System-FilePath.html
* https://hackage.haskell.org/package/filepath-1.5.4.0/docs/System-OsPath.html

And:

* https://hackage.haskell.org/package/directory-1.3.9.0/docs/System-Directory.html
* https://hackage.haskell.org/package/directory-1.3.9.0/docs/System-Directory-OsPath.html

tomjaguarpaw: Oh, OK, one of them re-exports stuff from `System.Environment`. Why?

hasufell: To have symmetric API. Otherwise you can only get those functions from process if you use OsString.

tomjaguarpaw: I don't understand why one would want to get them from `process`. They're already in `base`.

hasufell: There is no hard technical reason.

I think re-export makes sense, unless it's already in Prelude.

tbidne: As hasufell mentioned, I did this to be consistent with the other libraries that offer `OsPath/OsString` versions of an existing `FilePath` API (see also [path](https://hackage.haskell.org/package/path)).

However, it's fair to say that the situation here is slightly different, since those libraries merely translated an API that already existed within the library, whereas here we are adding an API that is new for the library itself.

Personally, I am indifferent on including a `FilePath` version, as I don't plan to use it myself. Happy to go with consensus.

changelog.md
tbidne: Good catch, thanks.

process.cabal
tomjaguarpaw: It's not a key part of the PR and it's easier to add it later than it is to remove it, so let's be cautious and omit it for now.

tomjaguarpaw:
Alright, this is looking good! @tbidne, is this ready to merge in your opinion?

Regarding testing more in the CI matrix, I'm happy for that to be added at a later date.

tbidne:
LGTM.

tomjaguarpaw (approved):

tomjaguarpaw:
Released as https://hackage.haskell.org/package/process-1.6.26.0

Thanks @tbidne and @hasufell

hasufell:
I didn't have time to review this in-depth but I'm not sure you're aware that `getArgs` on windows also includes the program name at position 0? The [documentation](https://hackage-content.haskell.org/package/Win32-2.14.2.2/docs/System-Win32-WindowsString-Console.html#v:getArgs) hints at it, but could be more explicit maybe.

See this as an example:

```
ghci> System.Environment.getArgs
[]
ghci> System.Process.Environment.OsString.getArgs
["C:\\\\ghcup\\\\ghc\\\\9.6.7\\\\bin/ghc-9.6.7.exe","--interactive","-package-env=-","-fbuilding-cabal-package","-O0","-outputdir","dist-newstyle\\\\build\\\\x86_64-windows\\\\ghc-9.6.7\\\\fake-package-0\\\\build","-odir","dist-newstyle\\\\build\\\\x86_64-windows\\\\ghc-9.6.7\\\\fake-package-0\\\\build","-hidir","dist-newstyle\\\\build\\\\x86_64-windows\\\\ghc-9.6.7\\\\fake-package-0\\\\build","-hiedir","dist-newstyle\\\\build\\\\x86_64-windows\\\\ghc-9.6.7\\\\fake-package-0\\\\build\\\\extra-compilation-artifacts\\\\hie","-stubdir","dist-newstyle\\\\build\\\\x86_64-windows\\\\ghc-9.6.7\\\\fake-package-0\\\\build","-i","-i.","-idist-newstyle\\\\build\\\\x86_64-windows\\\\ghc-9.6.7\\\\fake-package-0\\\\build","-idist-newstyle\\\\build\\\\x86_64-windows\\\\ghc-9.6.7\\\\fake-package-0\\\\build\\\\autogen","-idist-newstyle\\\\build\\\\x86_64-windows\\\\ghc-9.6.7\\\\fake-package-0\\\\build\\\\global-autogen","-Idist-newstyle\\\\build\\\\x86_64-windows\\\\ghc-9.6.7\\\\fake-package-0\\\\build\\\\autogen","-Idist-newstyle\\\\build\\\\x86_64-windows\\\\ghc-9.6.7\\\\fake-package-0\\\\build\\\\global-autogen","-Idist-newstyle\\\\build\\\\x86_64-windows\\\\ghc-9.6.7\\\\fake-package-0\\\\build","-optP-include","-optPdist-newstyle\\\\build\\\\x86_64-windows\\\\ghc-9.6.7\\\\fake-package-0\\\\build\\\\autogen\\\\cabal_macros.h","-this-unit-id","fake-package-0-inplace","-hide-all-packages","-Wmissing-home-modules","-no-user-package-db","-package-db","C:\\\\cabal\\\\store\\\\ghc-9.6.7\\\\package.db","-package-db","C:\\\\ghcup\\\\msys64\\\\tmp\\\\cabal-repl.-7680\\\\dist-newstyle\\\\packagedb\\\\ghc-9.6.7","-package-db","dist-newstyle\\\\build\\\\x86_64-windows\\\\ghc-9.6.7\\\\fake-package-0\\\\package.conf.inplace","-package-id","base-4.18.3.0","-package-id","process-1.6.26.0-118aa5a7b872de6fcb6fb40f642f211bc1820d2e","-XHaskell2010","-ghci-scriptC:\\\\ghcup\\\\msys64\\\\tmp\\\\cabal-repl.-7680\\\\setcwd.ghci","-package-id","process-1.6.26.0-118aa5a7b872de6fcb6fb40f642f211bc1820d2e","-package-id","Win32-2.14.2.1-fde6024115711a7c8c1b86f3c3eec27551e0c23a","-package-id","ghc-bignum-1.3","-package-id","ghc-prim-0.10.0","-package-id","rts-1.0.2","-package-id","filepath-1.4.301.0","-package-id","bytestring-0.11.5.4","-package-id","deepseq-1.4.8.1","-package-id","array-0.5.8.0","-package-id","template-haskell-2.20.0.0","-package-id","ghc-boot-th-9.6.7","-package-id","pretty-1.1.3.6","-package-id","exceptions-0.10.7","-package-id","mtl-2.3.1","-package-id","transformers-0.6.1.0","-package-id","stm-2.5.1.0","-package-id","containers-0.6.7","-package-id","directory-1.3.8.5","-package-id","Win32-2.13.3.0","-package-id","time-1.12.2","-package-id","process-1.6.19.0","-package-id","directory-1.3.10.1-8026e377ce9c3d47f1b21ce086f40562f1ab9698","-package-id","file-io-0.1.6-580273a821a8496f880f3ba8dd83e18afbc79c23","-package-id","time-1.14-84ccf33b0d93add4256e2b61b703c710e443120e","-hide-all-packages"]
```

tbidne:
I'm vaguely aware of subtleties regarding `argv[0]` and other issues e.g. https://github.com/haskell/cabal/issues/10275. But no, I was not aware of the windows/unix difference here.

What do you think should happen? Clarifying docs?

hasufell:
I'm not sure how to provide a consistent API to be honest. As you can see here it also shows all the arguments to `ghc.exe`. It behaves quite different than on unix. It might be the best to create a test suite with different ways of programs being executed to narrow down the exact behavior.

tbidne:
Oh you're right. This is quite unfortunate. To be explicit, the following program:

```haskell
import System.Environment qualified as Base
import System.Process.Environment.OsString qualified as Process

main :: IO ()
main = do
  Base.getArgs >>= p "Base: "
  Process.getArgs >>= p "Process: "
  where
    p h = putStrLn . (h ++) . show
```

Prints the following, per OS:

```
# osx
Base: ["arg1","arg2"]
Process: ["arg1","arg2"]

# ubuntu
Base: ["arg1","arg2"]
Process: ["arg1","arg2"]

# windows
Base: ["arg1","arg2"]
Process: ["D:\\a\\get-args-test\\get-args-test\\dist-newstyle\\build\\x86_64-windows\\ghc-9.14.1\\get-args-test-0.1\\x\\get-args-test\\opt\\build\\get-args-test\\get-args-test","arg1","arg2"]
```
Tested [here](https://github.com/tbidne/get-args-test).

A [stackoverflow](https://stackoverflow.com/a/60991478) answer suggested using [PathGetArgsW](https://learn.microsoft.com/hu-hu/windows/win32/api/shlwapi/nf-shlwapi-pathgetargsw) to strip the exe name. My naive test worked i.e. I forked `Win32` and added:

```haskell
foreign import WINDOWS_CCONV unsafe "shlwapi.h PathGetArgsW"
        pathGetArgsW :: LPCWSTR -> IO LPWSTR

-- normal getArgs for reference
getArgs :: IO [String]
getArgs = do
  getCommandLineW >>= peekTString >>= commandLineToArgv

getArgsNoExe :: IO [String]
getArgsNoExe = do
  getCommandLineW >>= pathGetArgsW >>= peekTString >>= commandLineToArgv
```

This produced:

```
Win32 getArgs: ["D:\\a\\get-args-test\\get-args-test\\dist-newstyle\\build\\x86_64-windows\\ghc-9.14.1\\get-args-test-0.1\\x\\get-args-test\\opt\\build\\get-args-test\\get-args-test","arg1","arg2"]
Win32 getArgsNoExe: ["arg1","arg2"]
```

If you think this is a promising direction, my tentative suggestion would be add this function to `Win32` (or change the existing one?), then have `process` use it instead. I can open a PR if so.

hasufell:
Did you test the ghci example? It's not just about the original arguments. There's also something about how "exec" works.

tbidne:
I'm afraid I did not. I don't have a windows machine, so I am relying on github's CI. I don't think `PathGetArgsW` would produce the same behavior there. All it does is drop the exe name from "correctly formatted" arg strings (e.g. whitespace quoted). Presumably you'd still have the rest of the `"--interactive" ...` args.

The ghci behavior seems bizarre. I'm not sure what to make of it.

tbidne:
Some context, GHC/base does some fairly sophisticated processing. On startup:

- Prepends the program name to `argv`: [link](https://gitlab.haskell.org/ghc/ghc/-/blob/41a19379fa97bc6c944526814fd644e6e54b1fed/rts/RtsFlags.c#L2654)
- Strips RTS args: [link](https://gitlab.haskell.org/ghc/ghc/-/blob/41a19379fa97bc6c944526814fd644e6e54b1fed/rts/RtsFlags.c#L788)

Then `getArgs`:

- Unconditionally drops the first arg: [link](https://gitlab.haskell.org/ghc/ghc/-/blob/41a19379fa97bc6c944526814fd644e6e54b1fed/libraries/ghc-internal/src/GHC/Internal/System/Environment.hs#L84)

Note that GHC also provides [`getFullArgs`](https://hackage-content.haskell.org/package/base-4.22.0.0/docs/GHC-Environment.html#v:getFullArgs). This does less processing on startup: [link](https://gitlab.haskell.org/ghc/ghc/-/blob/41a19379fa97bc6c944526814fd644e6e54b1fed/rts/RtsStartup.c#L292).

The impl in [unix](https://gitlab.haskell.org/ghc/packages/unix/-/blob/303cb8281dbc50f62461c2238667018c43debb42/System/Posix/Env/PosixString.hsc#L179) is basically the same as `base`, relying on the heavily processed `argv`. Naively, it seems like the simplest path to parity would be to do what `unix` does, use the heavily processed `base` version, but instead of a final

```haskell
CString -> ShortByteString -> PosixString
```

You'd have

```haskell
CString -> ShortByteString -> WindowsString
```
    
But maybe this won't work for some reason.

Edit: I guess this won't work because by the time we have `CString` it is too late. We'd need something like `CWString` or `Ptr Word16`. I can only think of two ways to attempt parity:

1. Attempt to replicate the same GHC parsing logic in "user space" i.e. `Win32` parses the prog name, drops it, and drops `RTS` args. Upside is no GHC changes, downside is ensuring consistency between e.g. `getArgs`, `getProgName`, `withArgs`...

2. GHC gets new `prog_win_argc`, `prog_win_argv`, duplicate the current `prog_argc`, `prog_argv` logic _without_ doing any decoding (see `RtsMain.c.getUTF8Args` and `Note [Windows Unicode Arguments]` in `RtsFlags.c`). The upside is that the API is much easier to implement. Downside is it means GHC changes, and windows RTS would be carrying two sets of program args.

hasufell:
https://github.com/ghc/ghc/blob/674858e3fb1b7fe644d7ec5d4b0c6d9f18d7bc51/driver/utils/cwrapper.c#L73-L161

This is what gets called when you invoke `ghci.exe`. So it is essentially `CreateProcess`. And from what I understand, there's no `exec` on windows, so we have to wait for the subprocess.

I'm honestly not sure what the expected behavior here would be. Why do you not see any arguments in ghci when you run `ghc --interactive` on unix and then type `getArgs` in the prompt?

hasufell:
GHCi changes both progname and arguments for evaluated expressions:

* https://github.com/ghc/ghc/blob/674858e3fb1b7fe644d7ec5d4b0c6d9f18d7bc51/ghc/GHCi/UI.hs#L542-L544

It then uses `withArgs`, which ends up calling the rts function `setProgArgv`: https://github.com/ghc/ghc/blob/674858e3fb1b7fe644d7ec5d4b0c6d9f18d7bc51/rts/RtsFlags.c#L2691

The Win32 functions do not follow these codepaths, so we broke `withArgs`, leading to this special case in ghci (and possibly other libraries that use it).

PR #346, opened 2025-06-22
Fix 'since' tags in System.Process.Environment.OsString

0rphee:
tags

### DIFF
--- a/System/Process/Environment/OsString.hs
+++ b/System/Process/Environment/OsString.hs
@@ -3,7 +3,7 @@
 
 -- | Miscellaneous information about the system environment, for 'OsString'.
 --
--- @since X.Y.Z
+-- @since 1.6.26.0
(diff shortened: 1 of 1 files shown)


tomjaguarpaw:
Oh, thanks!

tomjaguarpaw:
Does anyone understand the build error? I don't think it can have anything to do with this PR.

bgamari:
The problem appears to be that `REPLACEFILE_IGNORE_ACL_ERRORS` is [only defined](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-replacefilea) when `_WIN32_WINNT >= 0x0600`. I don't honestly know why we are only seeing this now but it does seem like a real bug in `Win32`.

tomjaguarpaw:
Looks like CI is clear now. Thanks for chasing that up @bgamari.

PR #347, opened 2025-08-17
Make command line escaping on Windows not add superfluous double quotes

### LINKED ISSUE #51: Quotation in System.Process.system for Windows
Moved from https://ghc.haskell.org/trac/ghc/ticket/5376:

Wagnerdm writes: I had a bit of fun recently tracking down quoting issues with the "system" command in Windows. For the examples below, I'll consistently use "Windows> " as the beginning of some text sent to the Windows command prompt cmd.exe, and use "GHC> " as the beginning of some text sent to a ghci session running in cmd.exe with System.Cmd imported.

The situation is this: I want to hand off a command line which has both a possibly-quoted command name and a (definitely) quoted argument. For concreteness, let's use "more" as the command and "foo.txt" as the argument, so that you can follow along at home on your favorite Windows system.

```
Windows> echo foo > foo.txt
Windows> "more" "foo.txt"
foo
```
All good so far. But:

```
GHC> system "\"more\" \"foo.txt\""
'more" "foo.txt' is not recognized as an internal or external command, operable program or batch file.
ExitFailure 1
```
After some digging, I discovered that system is shipping out to cmd /c, and indeed:

```
Windows> cmd /c "more" "foo.txt"
'more" "foo.txt' is not recognized as an internal or external command, operable program or batch file.
```
I don't know what the *right* fix is. However, after a bit of playing around, I discovered the following:

```
Windows> cmd /c ""more" "foo.txt""
foo
GHC> system "\"\"more\" \"foo.txt\"\""
foo
ExitSuccess
```
Wrapping commands with an extra pair of double-quotes this way seemed to give behavior matching the bare cmd.exe for all the examples I could think of, even ones I thought it would break. For example:
```
GHC> system "\"more foo.txt\""
foo
ExitSuccess
```
If this turns out to be the right thing to do, it's pretty easy to implement. In the commandToProcess function, at libraries/process/System/Process/Internals.hs:455, the change is just

```
-   return (cmd, translate cmd ++ "/c " ++ string)
+   return (cmd, translate cmd ++ "/c \"" ++ string ++ "\"")
```
(And in any case, the examples above should answer this amusing comment, immediately following those lines:
```
	-- We don't want to put the cmd into a single
	-- argument, because cmd.exe will not try to split it up.  Instead,
	-- we just tack the command on the end of the cmd.exe command line,
	-- which partly works.  There seem to be some quoting issues, but
	-- I don't have the energy to find+fix them right now (ToDo). --SDM
	-- (later) Now I don't know what the above comment means.  sigh.
```
Later, Brandon comments: What that comment means is that how CMD.EXE handles spaces is Windows-version-dependent. On the other hand, I think it's mostly consistent between XP and Windows 7 --- and I feel sorry for anyone forced to use an older version.

thomasjm:
This PR is an attempt to fix #51. 

As explained there, `System.Process` currently adds superfluous double quotes when assembling command lines on Windows. For example, if I run `readCreateProcess (proc "notepad.exe" ["foo.txt"]) ""`, then Windows will see a command line like `"notepad.exe" "foo.txt"`.

This quoting is extra-conservative, but it also causes a number of problems which people have reported in #51. In my case, it makes it impossible to run something like `readCreateProcess (proc "wsl.exe" ["--help"]) "`, because the quotes confuse `wsl.exe` and it tries to go into shell mode and execute commands on a VM. The quoting is also contrary to what other common process libraries do, like Python's `subprocess.run`.

## Correctness

To show that this alternative implementation is correct, I wrote some tests in this repo: https://github.com/thomasjm/process-escape-windows. I test in the following ways:

* Using `CommandLineToArgvW` FFI. Using FFI, I assemble random executable names and argument strings and test that the Windows `CommandLineToArgvW` function parses them correctly (with QuickCheck).
* Running `cmd.exe /c child.exe <args>`, where `child.exe` is a specially crafted C program that echoes back its arguments. We have to apply an extra level of escaping for `cmd.exe` for this to work. This is also tested using QuickCheck.
* I also tested the original `translateInternal` from this repo using FFI.

Conclusions:

* The current implementation in this package is correct for normal programs! (But not for `.bat` files; see below)
* The new version in this PR is also correct for normal programs.

## About `.bat` and `.cmd` files

This package also has a function called `translateCmdExeArg`, which is applied when you use `RawCommand` and an executable name that ends with `.bat` or `.cmd`.

I tried for a while to make a version of `translateCmdExeArg` that will work for arbitrary arguments, tested with QuickCheck--but ultimately succumbed to despair. Look at how long and crazy a typical StackOverflow answer is on this subject: https://stackoverflow.com/a/31413730.

For this reason, I haven't touched `translateCmdExeArg` in this PR. I immediately guessed that it would fail for arguments that contain double quotes. For example, I tested running `readCreateProcess (proc childBat ["a\"b"]) ""` using a recent Stackage resolver, using `child.bat` in my test repo, which echoes back the commands it receives. It left a random caret in the result: `a"b^`.

It's good that `translateCmdExeArg` has a [BatBadBut](https://flatt.tech/research/posts/batbadbut-you-cant-securely-execute-commands-on-windows/) remediation, but I would say users shouldn't expect it to pass their arguments faithfully through to a batch script if the arguments have unusual characters etc.

## Conclusion

Hopefully this is enough validation to support merging this PR! We could also potentially bring my tests over to this package, but that would involve adding a `QuickCheck` dependency.

### DIFF
--- a/System/Process/Windows.hsc
+++ b/System/Process/Windows.hsc
@@ -422,18 +422,21 @@ commandToProcess
   -> IO (FilePath, String)
 commandToProcess (ShellCommand string) = do
   cmd <- findCommandInterpreter
-  return (cmd, translateInternal cmd ++ " /c " ++ string)
-        -- We don't want to put the cmd into a single
-        -- argument, because cmd.exe will not try to split it up.  Instead,
-        -- we just tack the command on the end of the cmd.exe command line,
-        -- which partly works.  There seem to be some quoting issues, but
-        -- I don't have the energy to find+fix them right now (ToDo). --SDM
-        -- (later) Now I don't know what the above comment means.  sigh.
+  -- Note: this is a way to pass a command directly to cmd.exe. Callers are
+  -- responsible for properly quoting and sanitizing this string.
+  return (cmd, translateInternal0 cmd ++ " /c " ++ string)
 commandToProcess (RawCommand cmd args)
   | map toLower (takeWinExtension cmd) `elem` [".bat", ".cmd"]
-  = return (cmd, translateInternal cmd ++ concatMap ((' ':) . translateCmdExeArg) args)
+  = return (cmd, translateInternal0 cmd ++ concatMap ((' ':) . translateCmdExeArg) args)
   | otherwise
-  = return (cmd, translateInternal cmd ++ concatMap ((' ':) . translateInternal) args)
+  -- Note: on Windows, commands are passed as a single string of space-delimited
+  -- pieces, *not* as an executable name + list of args as on POSIX. Windows
+  -- programs can inspect this command string and parse it however they want.
+  --
+  -- However, most programs use the CommandLineToArgvW function from the Win32
+  -- API to parse their arguments. Here we escape the argument in such a way
+  -- that they'll always come through correctly with this function.
+  = return (cmd, translateInternal0 cmd ++ concatMap ((' ':) . translateInternal) args)
 
 -- TODO: filepath should also be updated with 'takeWinExtension'. Perhaps
 -- some day we can remove this logic from `process` but there is no hurry.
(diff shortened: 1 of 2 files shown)


tomjaguarpaw:
Thanks for this. I don't personally have any knowledge or understanding of the Windows process API or access to a Windows system. Maybe one of my `process` maintainer collaborators can help with this one.

PR #348, opened 2025-09-12
Switch to the new operations for obtaining OS handles

jeltsch:
There is an effort to move some GHC-specific modules out of `base`. Candidates for such moving are `GHC.IO.Handle.Types` and `GHC.IO.Handle.Internals`. These modules are often used for obtaining operating-system handles (file descriptors, Windows handles) from Haskell handles. Such uses are also present in the `process` package.

There is a proposal to add dedicated operations for operating-system handle acquisition to `base`. A corresponding draft implementation can be found in GHC merge request [!14732](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/14732). The present pull request for `process` changes the package to use those new operations.

Note that this pull request changes the behavior of `System.Process.Common.mbFd` in two ways:
1. The given handle is released between setting it to “blocking” and acquiring its file descriptor. It should be checked whether this may cause problems when another thread concurrently accesses the handle.
2. In the case that the given handle is a duplex handle, *both* ends are set to “blocking” instead of just the reading end, as it was before. This was easier to implement based on the `withHandle` family of operations from `GHC.IO.Handle.Internals` but may not be what is desired.

### DIFF
--- a/System/Process/Common.hs
+++ b/System/Process/Common.hs
@@ -45,13 +45,16 @@ import Data.String ( IsString(..) )
 import Foreign.Ptr
 import Foreign.Storable ( Storable(peek) )
 
+import System.IO.OS (withFileDescriptorReadingBiasedRaw)
+#if defined(__IO_MANAGER_WINIO__)
+import System.IO.OS (withWindowsHandleReadingBiasedRaw)
+#endif
 import System.Posix.Internals
 import GHC.IO.Exception
 import GHC.IO.Encoding
 import qualified GHC.IO.FD as FD
 import GHC.IO.Device
 #if defined(__IO_MANAGER_WINIO__)
-import GHC.IO.Handle.Windows
 import GHC.IO.Windows.Handle (fromHANDLE, Io(), NativeHandle())
 #endif
 import GHC.IO.Handle.FD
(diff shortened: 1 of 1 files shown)


jeltsch:
Note that the tests necessarily fail with currently released GHC versions, as they require the new module `System.IO.OS`, which is added by GHC ticket [!14732](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/14732). This is also the reason why I gave this pull request draft status.

jeltsch:
I removed the draft status of this pull request for it not to get lost.

tomjaguarpaw:
Thank you very much for this contribution. `process` seems to be undermaintained at the moment. I'm a technically a maintainer, but I see myself as a "stop gap" or "emergency" maintainer, since I don't have the expertise necessary to make meaningful contributions to this library.

I will try to clarify exactly who the maintainers of `process` are, and see if I can find someone with the knowledge to review your PR.

Perhaps @jeltsch, you'd like to become a maintainer of `process`?

jeltsch:
Thanks a lot for moving this forward.

As for myself, I don’t see myself as a future maintainer of `process`, I’m afraid.

tomjaguarpaw:
@bgamari, do you have the capacity to take on the review of this?

PR #352, opened 2026-03-07
Release 1.6.26.2

tomjaguarpaw:

PR #353, opened 2026-03-07
Support Cabal 3.16 (only)

tomjaguarpaw:
The Cabal API broke in 3.16 so supporting 3.16 and older versions would require some form of conditional compilation

@bgamari , are you OK with this?  @sheaf, you wrote this, what do you think?

Github runners seem to have migrated to 3.16 and I can't merge https://github.com/haskell/process/pull/352 until we have a fix in place. I'm OK with finding some way (maybe CPP) of widening the bound again later. Since this is just for tests I don't think it's critical.

### DIFF
--- a/test/Setup.hs
+++ b/test/Setup.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP #-}
 {-# OPTIONS_GHC -Wall #-}
 
 module Main (main) where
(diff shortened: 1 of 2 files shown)


tomjaguarpaw:
Well, that's no good because 3.16 doesn't support the older GHCs we still support. I'll just do the CPP.

PR #355, opened 2026-03-08
Fix changelog; release 1.6.27.0

tomjaguarpaw:
1.6.26.2 was PVP-violating so will be marked as a deprecated version


tomjaguarpaw:
Minor version released here https://hackage.haskell.org/package/process-1.6.27.0

PR #358, opened 2026-04-14
Add `callCreateProcess`

sol:
- analog to `readCreateProcess`
- this also removes code duplication between `callProcess` and `callCommand`

### DIFF
--- a/System/Process.hs
+++ b/System/Process.hs
@@ -352,26 +352,15 @@ spawnCommand cmd = do
 callCreateProcess :: CreateProcess -> IO ()
 callCreateProcess = callCreateProcess_ "callCreateProcess"
 
--- | Creates a new process to run the specified command with the given
--- arguments, and wait for it to finish.  If the command returns a non-zero
--- exit code, an exception is raised.
---
--- If an asynchronous exception is thrown to the thread executing
--- @callProcess@, the forked process will be terminated and
--- @callProcess@ will wait (block) until the process has been
--- terminated.
+-- | \"@callProcess cmd args@\" is a shorthand for
+-- \"@'callCreateProcess' ('proc' cmd args)@\".
 --
 -- @since 1.2.0.0
 callProcess :: FilePath -> [String] -> IO ()
 callProcess cmd = callCreateProcess_ "callProcess" . proc cmd
 
--- | Creates a new process to run the specified shell command.  If the
--- command returns a non-zero exit code, an exception is raised.
---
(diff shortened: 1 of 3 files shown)


System/Process.hs
```diff
@@ -348,13 +363,7 @@ spawnCommand cmd = do
...
 -- @since 1.2.0.0
 callProcess :: FilePath -> [String] -> IO ()
-callProcess cmd args = do
-    exit_code <- withCreateProcess_ "callProcess"
-                   (proc cmd args) { delegate_ctlc = True } $ \_ _ _ p ->
-                   waitForProcess p
-    case exit_code of
-      ExitSuccess   -> return ()
-      ExitFailure r -> processFailedException "callProcess" cmd args r
+callProcess cmd = callCreateProcess_ "callProcess" . proc cmd
```
sol, line 360: Technically I would be in favor of:

1. Implementing this in terms of  `callCreateProcess`
1. get rid of  `callCreateProcess_`
1. state in the documentation that this is an alias for `callCreateProcess (proc cmd args)`

However, technically that is a breaking change, as it will change error messages.  Not sure how big of an issue that is, but when `readCreateProcess` was introduced back in 2015 error messages did change, so there is some precedence at least.

From a users perspective, you don't really want a whole bunch of different functions that are all slightly different and are implemented in terms of internal primitives.  What you want is:

1. General functions that are flexible
2. Specialized functions that are implemented in terms of those more general functions

Sadly, `process` is very far from this ideal.

tomjaguarpaw:
Thanks, this is very nice. I took the liberty of breaking into tiny commits so I could see what was going on.

System/Process.hs
tomjaguarpaw: And presumably for `callCommand` too?

I don't really have an objection to this. It makes a lot of sense and I personally don't feel strongly about just the function name changing in the error message.

tomjaguarpaw: Looks like we need `spawnCreateProcess` too!

tomjaguarpaw:
There are failures only on old GHCs on MacOS. At some point I might dig into why, but they shouldn't block this PR.

System/Process.hs
tomjaguarpaw: For the latter see https://github.com/haskell/process/issues/360

tomjaguarpaw:
This is released here: https://hackage-content.haskell.org/package/process-1.6.28.0/docs/System-Process.html#v:callCreateProcess

PR #364, opened 2026-04-28
Explain autoreconf -i in README

tomjaguarpaw:


bgamari:
On one hand, this is useful to articulate given the status quo.

On the other hand, I think the current situation has shown that the status quo is somewhat untenable. We should at very least be intentional about the version and provenance of `autotools` used to generate the configure script. More broadly, I suspect a more automated process is in order.

tomjaguarpaw:
> I think the current situation has shown that the status quo is somewhat untenable. We should at very least be intentional about the version and provenance of autotools used to generate the configure script. More broadly, I suspect a more automated process is in order.

Totally agree!

I suggest we merge this for better knowledge-sharing pending development of a better release process.

bgamari:
As well, it is in general a good practice to `git clean -fxd` prior to producing a source distribution to ensure that things like stale `configure` are caught. Perhaps you could mention this here as well?

tomjaguarpaw:
OK, done.

tomjaguarpaw:
I intend to merge this is due course @bgamari, unless you have any further comments.

bgamari-positron (approved):

PR #366, opened 2026-04-29
Stub out runProcess.c on wasm32

### LINKED ISSUE #357: On process 1.6.27.0, sys/wait.h is included unconditionally which breaks builds on wasm32-wasi
Commit 5c98f59 moved the `#if defined(HAVE_FORK)` guard in cbits/posix/runProcess.c but left
`#include <sys/wait.h>` outside the guard. 

On wasm32-wasi, sys/wait.h does not exist and HAVE_FORK is not defined, causing a build failure.

This was found on the x86_64-linux-alpine3_23-wasm-cross_wasm32-wasi-release GHC CI job.

wz1000:
Also bump base upper bound to < 4.24 for GHC 10.0

I can open a new MR for the base bound bump, but since we don't have CI setup
for wasm32-wasi, it was more convienent to have both changes in a single branch so that
this change can be tested on GHC CI: https://gitlab.haskell.org/ghc/ghc/-/merge_requests/15978

Fixes #357

### DIFF
--- a/cbits/posix/runProcess.c
+++ b/cbits/posix/runProcess.c
@@ -5,6 +5,11 @@
    ------------------------------------------------------------------------- */
 
 #include "runProcess.h"
+
+// Callers are stubbed out on wasm32: runInteractiveProcess (System/Process/Posix.hs),
+// terminateProcess, getProcessExitCode, waitForProcess (System/Process.hs).
+#if !defined(wasm32_HOST_ARCH)
+
 #include "common.h"
 
 #include <unistd.h>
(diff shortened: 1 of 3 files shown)


MangoIV:
Looks good to me. 
I think I would like this commit message

> The Haskell side callers of this file are already stubbed out on this platform.

adjusted with a permalink to the respective haskell side caller.

tomjaguarpaw:
Hi @wz1000, what do you think of @MangoIV's idea, or at least leaving a comment explaining why this is disabled on wasm32?

wz1000:
I have added a comment and amended the commit message.

MangoIV:
Neat, thanks @wz1000, looks good to me then.

tomjaguarpaw:
Thanks, will merge this soon unless @bgamari objects

PR #369, opened 2026-05-04
Add a basic editorconfig file. 

MangoIV:
Editorconfig can be used to unintrusevely make editors cohere with basic formatting standards used in a projects, e.g. 
- tabs vs spaces 
- default tab width (i.e. when doing line breaks) 
- trailing whitespace 
- file encoding

It is supported by all major editors and doesn't need any extra configuration in most. 

Vscode and neovim for example just make it work out of the box.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_style = space
+insert_final_newline = true
+tab_width = 2
+trim_trailing_whitespace = true


tomjaguarpaw:
Seems OK to me. @bgamari?

Is there also a command line tool to apply the formatting, in case someone is using an editor that doesn't support it? (I have no idea how I would get this to work on Emacs, for example.)

MangoIV:
yes, I think there's a command line tool, but I think the main advantage is that it catches "the vast majority" of cases without being otherwise intrusive. As you see I didn't reformat the project either, even though there might e.g. be trailing white space or tabs elsewhere. 

Wrt emacs, it should be supported "out of the box" afaict [link to docs](https://www.gnu.org/software/emacs/manual/html_node/emacs/EditorConfig-support.html)

tomjaguarpaw:
Will merge this soon unless @bgamari objects

PR #375, opened 2026-06-09
fix: move check for posix_spawn_file_actions{,_np} to runtime

MangoIV:
to avoid issues with symbols not being available on macos, we move the check for the existance of posix_spawn_file_actions_addchdir{,np} entirely to the runtime.

The gist is that when testing the RC for 9.12.5, I found another issue with symbols not being there when they were configured to be. To avoid this issue altogether I moved the check to the runtime by first creating a weak symbol for the two functions and then the linker will overwrite it if available.

### DIFF
--- a/cbits/posix/posix_spawn.c
+++ b/cbits/posix/posix_spawn.c
@@ -93,6 +93,43 @@ setup_std_handle_spawn (int fd,
     }
 }
 
+
+int __attribute__((weak)) posix_spawn_file_actions_addchdir
+  (posix_spawn_file_actions_t *restrict file_actions, const char *restrict path);
+int __attribute__((weak)) posix_spawn_file_actions_addchdir_np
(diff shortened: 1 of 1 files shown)


tomjaguarpaw:
Can you please let me know the timeline that this is needed in? Is this a blocker for a GHC release?

MangoIV:
@tomjaguarpaw yes, so it’s rather urgent but it’s not ready yet, I have to do a bit more investigation

tomjaguarpaw:
OK, please ping me when you have something urgent I need to do.

MangoIV:
I still think ideally we should move all the checking to runtime since configure might give us all sorts of non-sense (cf. the recent macos issues) and is in comparison "cheap".

MangoIV:
E.g. I'm dreading the moment apple will remove the `np` version of the `posix_spawn_file_actions` symbols which is deprecated rn.

PR #377, opened 2026-06-10
darwin: Allow posix_spawn_file_actions_addchdir to be missing when linking against process

### LINKED ISSUE #376: Fix for #356 is incomplete, resulting in link errors on macos on certain configurations.
https://github.com/haskell/process/pull/363 attempted to fix https://github.com/haskell/process/issues/356

`posix_spawn_file_actions_addchdir` was introduced in macOS 26.

When `MACOSX_DEPLOYMENT_TARGET=11.0` (as it is on GHC CI) is set, clang will emit a weak reference to this symbol. However, weak references on MachO objects don't behave like weak references on ELF (see https://github.com/iains/gcc-darwin-arm64/discussions/133). A final link with undefined weak references still fails.

We see this on GHC CI, where a process library built against macOS 26 is attempted to be linked against a homebrew/clang toolchain not exposing `posix_spawn_file_actions_addchdir`, resulting in a link failure: 
https://gitlab.haskell.org/ghc/ghcup-ci/-/jobs/2539703

```
Undefined symbols for architecture arm64:
  "_posix_spawn_file_actions_addchdir", referenced from:
      _do_spawn_posix in libHSprocess-1.6.29.0-d296.a[4](posix_spawn.o)
      _do_spawn_posix in libHSprocess-1.6.29.0-d296.a[4](posix_spawn.o)
```

I think that the correct fix is to pass `-Wl,-U,_posix_spawn_file_actions_addchdir` in `ld-options:` in the cabal file, which will allow the link to succeed when this symbol is missing (setting it to null), which will then allow the fix in #363 to take effect.

wz1000:
In 0699e410418f5fe8eaf164ce21048b8b25c9b112 we attempted to account for darwin toolchains missing this symbol, which is imported as a weak references.

However, weak references on MachO objects don't work the same way as on ELF (see https://github.com/iains/gcc-darwin-arm64/discussions/133). A final link with undefined weak references still fails.

To fix this, on darwin, we explicilty add `-Wl,-U,_posix_spawn_file_actions_addchdir` to `ld-options` so anything linking against process explicilty instructs the linker to ignore undefined references to this symbol (set it to NULL at runtime).

Fixes #376

### DIFF
--- a/process.cabal
+++ b/process.cabal
@@ -87,6 +87,19 @@ library
                 cbits/posix/find_executable.c
             other-modules: System.Process.Posix
 
+    -- On macOS, posix_spawn_file_actions_addchdir is available from macOS >= 26.
+    -- We weakly reference this symbol when we are build on such a platform.
+    -- However, weak references on MachO are hints to the dynamic loader, not the
+    -- static linker, which will still expect the symbol to be defined somewhere.
+    -- This means linking fails when we try to link an executable depending on
+    -- process on a toolchain which doesn't have this symbol against a process
+    -- library build on a toolchain with the symbol.
+    -- See https://github.com/haskell/process/issues/376
+    -- Hence: instruct the linker to explicilty allow undefined references to
+    -- the symbol when linking against process on darwin
+    if os(darwin)
+        ld-options: -Wl,-U,_posix_spawn_file_actions_addchdir
+
     include-dirs: include
     install-includes:
         runProcess.h


tomjaguarpaw:
@MangoIV, can you comment on this?

wz1000:
This should supersede https://github.com/haskell/process/pull/375, but we will validate this against GHC CI and get back to you.

MangoIV:
I think this is a good fix. Since my solution doesn’t seem to work on its own, either, I think this is good. I wonder whether we should do both, though. Probably not right now though since this is the minimal fix

tomjaguarpaw:
OK, and this is urgent due to impending GHC release, right?

MangoIV:
yes, this is blocking the upcoming GHC releases.

tomjaguarpaw (approved):

PR #380, opened 2026-07-21
do_spawn_fork: Check more syscalls for errors

arrowd:
This might help diagnosing https://gitlab.haskell.org/ghc/ghc/-/work_items/27512


tomjaguarpaw:
Thanks! Looks reasonable to me, but I'm not an expert. @bgamari?

Issue #381, opened 2026-08-31
Process creation on Windows is needlessly serialised


sheaf:
When `close_fds = True`, we can explicitly pass the processes that the child is allowed to inherit. This completely avoids the problem of [GHC bug #2650](https://gitlab.haskell.org/ghc/ghc/-/work_items/2650) that made locking necessary.

In a microbenchmark, spawning 200 processes with `close_fds = True` concurrently takes about 1.5s on my machine, with over 40% of the time (0.7s) wasted due to waiting on the lock.

While working on speeding up GHC's downsweep pass, I found that on Windows a lot of time was spent on process spawning (e.g. to run the C preprocessor), and that this serialisation was a significant component of that. In a rebuild of the `vulkan` package, the downsweep pass spawns a huge amount of subprocesses, and removing this locking takes the time spent in downsweep from 31s to 27s (4s total timesave).

PR #382, opened 2026-08-31
Windows: improve "close_fds=True" spawning

sheaf:
Ticket: #381.

Two fixes to process spawning with "close_fds = True" on Windows:

  1. Use PROC_THREAD_ATTRIBUTE_HANDLE_LIST to avoid inheriting handles from the parent.
  2. Allow concurrent spawns of subprocesses with "close_fds = True".

### DIFF
--- a/System/Process/Windows.hsc
+++ b/System/Process/Windows.hsc
@@ -130,17 +130,7 @@ createProcess_Internal_mio fun def@CreateProcess{
        fdout <- mbFd fun fd_stdout mb_stdout
        fderr <- mbFd fun fd_stderr mb_stderr
 
-       -- #2650: we must ensure mutual exclusion of c_runInteractiveProcess,
-       -- because otherwise there is a race condition whereby one thread
-       -- has created some pipes, and another thread spawns a process which
-       -- accidentally inherits some of the pipe handles that the first
-       -- thread has created.
-       --
-       -- An MVar in Haskell is the best way to do this, because there
-       -- is no way to do one-time thread-safe initialisation of a mutex
-       -- the C code.  Also the MVar will be cheaper when not running
-       -- the threaded RTS.
-       proc_handle <- withMVar runInteractiveProcess_lock $ \_ ->
+       proc_handle <- withSpawnLock mb_close_fds $
                       throwErrnoIfBadPHandle fun $
                            c_runInteractiveProcess pcmdline pWorkDir pEnv
                                   fdin fdout fderr
@@ -229,17 +219,7 @@ createProcess_Internal_winio fun def@CreateProcess{
      hwnd_out <- mbHANDLE _stdout mb_stdout
      hwnd_err <- mbHANDLE _stderr mb_stderr
 
-     -- #2650: we must ensure mutual exclusion of c_runInteractiveProcess,
-     -- because otherwise there is a race condition whereby one thread
-     -- has created some pipes, and another thread spawns a process which
-     -- accidentally inherits some of the pipe handles that the first
-     -- thread has created.
-     --
-     -- An MVar in Haskell is the best way to do this, because there
-     -- is no way to do one-time thread-safe initialisation of a mutex
-     -- the C code.  Also the MVar will be cheaper when not running
-     -- the threaded RTS.
-     proc_handle <- withMVar runInteractiveProcess_lock $ \_ ->
+     proc_handle <- withSpawnLock mb_close_fds $
                     throwErrnoIfBadPHandle fun $
                          c_runInteractiveProcessHANDLE pcmdline pWorkDir pEnv
                                 hwnd_in hwnd_out hwnd_err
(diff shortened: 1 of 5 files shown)


sheaf:
@Mistuke Perhaps you could take a look at this change? It helps avoid unnecessarily locking when spawning processes on Windows, which I measured to have a big impact on GHC when compiling packages with a lot of modules that use `CPP`.

System/Process/Windows.hsc
```diff
@@ -267,9 +247,94 @@ createProcess_Internal_winio fun def@CreateProcess{
...
+data SpawnLock =
+  SpawnLock
+    { spawn_lock :: !(MVar ())
+      -- ^ lock to hold for concurrent process creation
+    , shared_spawns_count :: !(MVar Int)
+      -- ^ how many @close_fds = True@ spawns are in flight
+    , spawn_turnstile :: !(MVar ())
+      -- ^ turnstile used to implement the fairness guarantee
+      -- in Note [Concurrent process spawning]
+    }
```
dcoutts, line 299: Correct me if I'm wrong, but isn't this exactly a reader/writer lock? Allow many readers (close_fds = True) **xor** one writer (close_fds = False), right?

Is there one available in the dependencies of `process` that we can reuse?

sheaf: I found two somewhat antiquated packages on Hackage that provide something of the sort:

  - [`Control.Concurrent.ReadWriteLock`](https://hackage.haskell.org/package/concurrent-extra/docs/Control-Concurrent-ReadWriteLock.html) from the `concurrent-extra` package.
  - [`Control.Concurrent.FairRWLock`](https://hackage.haskell.org/package/SafeSemaphore/docs/Control-Concurrent-FairRWLock.html) from the `SafeSemaphore` package.

Only the second implements the fairness guarantee we are after here.

`process` is a boot library with very few dependencies: only really `base` and `unix`/`Win32`. So I guess if we wanted this functionality anywhere it would have to be in `base`; a new library just for this doesnt seem worth it.

Mistuke:
Hiya, sorry for the quiet, my current workload means I can only work on GHC stuff on the weekends. I have some free time this Sunday so will review it then!

System/Process/Windows.hsc
Mistuke: Just a quick note, Windows has proper reader/writer locks primitives now and the minimal OS dependencies mandates they be there. The RTS depends on this.

I haven't looked at it closely yet, just making a quick comment.

cbits/win32/runProcess.c
```diff
@@ -6,6 +6,14 @@
 
 #define UNICODE
 
+/* We need _WIN32_WINNT >= 0x0600 for STARTUPINFOEX and PROC_THREAD_ATTRIBUTE_HANDLE_LIST,
+   while the MINGW toolchains bundled with older GHCs (prior to GHC 9.4) default
+   it to below that.  */
+#if !defined(_WIN32_WINNT) || _WIN32_WINNT < 0x0600
```
Mistuke, line 12: I'm surprised this is needed, since any program that goes through GHC as the C compiler should get `-D_WIN32_WINNT=0x06010000` set for both C and CPP.

But I guess with the package refactorings in GHC this may not be working, my suggestion if it is needed is to move this to the cabal file, and make it `0x06010000` to align with GHC.

Though this shouldn't be needed.

cbits/win32/runProcess.c
```diff
@@ -348,6 +361,53 @@ runInteractiveProcessWrapper (
...
+
+    if (close_fds && inherit && nInheritedHandles == 0) {
+        /* UpdateProcThreadAttribute requires a non-empty list of handles;
+           shortcut if there are no handles to inherit. */
+        inherit = FALSE;
+    } else if (nInheritedHandles > 0) {
+        SIZE_T attrSize = 0;
+        InitializeProcThreadAttributeList(NULL, 1, 0, &attrSize);
+        if (attrSize == 0) {
+            SetLastError(ERROR_INVALID_PARAMETER);
```
Mistuke, line 386: would it be better to propagate the actual error from `InitializeProcThreadAttributeList`?

cbits/win32/runProcess.c
```diff
@@ -348,6 +361,53 @@ runInteractiveProcessWrapper (
...
+            goto cleanup_err;
+        }
+        attrList = (LPPROC_THREAD_ATTRIBUTE_LIST) HeapAlloc(GetProcessHeap(), 0, attrSize);
+        if (attrList == NULL) {
+            SetLastError(ERROR_NOT_ENOUGH_MEMORY);
+            goto cleanup_err;
+        }
+        if (!InitializeProcThreadAttributeList(attrList, 1, 0, &attrSize)) {
+            DWORD err = GetLastError();
+            HeapFree(GetProcessHeap(), 0, attrList);
```
Mistuke, line 396: this is all done by `cleanup_err` no? so all you need is the goto right?

System/Process/Windows.hsc
```diff
@@ -267,9 +247,94 @@ createProcess_Internal_winio fun def@CreateProcess{
...
+by a collection of 'close_fds = True' spawns. We differentiate the two by a
+second 'MVar', 'shared_spawns_count', which counts how many joint
+'close_fds = True' spawns are currently underway.
+Fairness is implemented using a third 'MVar', the turnstile. A waiting exclusive
+spawn holds the turnstile 'MVar', so that further non-exclusive spawns are put
+on wait rather than accumulating with existing in-flight shared spawns, which
+could end up starving exclusive spawns.
+
+The solution is implemented on the Haskell side using 'MVar's, because there is
+no way to do one-time thread-safe initialisation of a mutex in the C code.
```
Mistuke, line 285: this comment is slightly out of date, since you can use a static constructor to initialize it.

System/Process/Windows.hsc
Mistuke: You can definitely just use a slim read/writer lock here https://learn.microsoft.com/en-us/windows/win32/Sync/slim-reader-writer--srw--locks and move the safety guarantee into `c_runInteractiveProcessHANDLE` which would be faster.

you can also just use static initialization through `SRWLOCK_INIT` which is intended for this.
so both these work and are thread safe

```
#include <windows.h>

static SRWLOCK g_lock = SRWLOCK_INIT;

void foo(void)
{
    AcquireSRWLockExclusive(&g_lock);

    /* protected work */

    ReleaseSRWLockExclusive(&g_lock);
}
```
and
```
#include <windows.h>

static SRWLOCK g_lock;

__attribute__((constructor))
static void init_globals(void)
{
    InitializeSRWLock(&g_lock);
}
```
but the former compatible with more compilers.

This is already an improvement but should you want to fix the API itself you can do the above.

Mistuke (commented):
looks pretty good to me! left some comments.

Sorry for the delay, I forgot, next time please do ping

cbits/win32/runProcess.c
sheaf: `-D_WIN32_WINNT=0x06010000` is set when compiling the RTS but nothing else (as far as I can tell).

I don't want to hard-code a value in `process.cabal` as that could go out of sync if someone updates the value used in the RTS but doesn't update `process`, while the CPP here only sets it as a fallback.

Given that `process` is supposed to work across a range of GHC versions I think the CPP as-written is appropriate.

System/Process/Windows.hsc
sheaf: Thanks for the suggestion, indeed that seems much better, significantly more straightforward. I have implemented that now, please take a look.

cbits/win32/runProcess.c
sheaf: The issue is that whether to call `DeleteProcThreadAttributeList` for cleanup depends on whether initialisation succeeded or not.

I refactored the handling of the attribute list to a separate function to make it a bit clearer.

sheaf (commented):
@Mistuke Thank you very much for the review. I moved the lock implementation into the C code using SRW locks, and tidied up the initialisation/cleanup logic of the attribute list. I would appreciate another look!

cbits/win32/runProcess.c
Mistuke: > -D_WIN32_WINNT=0x06010000 is set when compiling the RTS but nothing else (as far as I can tell).
So the original reason for this was that the RTS package was hardwired into every compilation.  We used to set this in `SysTools.hs` because ultimately the compiler has it's own minimum requirement, and the inclusion of `Rts.h` *requires* this minimum version.  At some point the RTS package was migrated to a cabal package but it's flags were still supposed to apply to all compilation.

Then we had `ghc-prim` where we set it as well to get all uses of `base`.  So I'm surprised it's not already passed, but I see `ghc-*` were refactored but I kinda still expected this invariant..

that said

> Given that process is supposed to work across a range of GHC versions I think the CPP as-written is appropriate.

The guard as is doesn't really give you anything on older GHC versions though. You'll still just get a link failure on undefined values.

Unless this has changed recently, the headers just use these are include guards, and on older headers the definitions just won't be there.  This is why `win32` explicitly adds `_compat.h` header files that move the missing values into the package https://github.com/haskell/win32/tree/master/include so that older compilers still work.

So I don't feel very strongly about this, but there's no additional safety here.

cbits/win32/runProcess.c
```diff
@@ -282,59 +290,153 @@ setStdHandleInfo (LPHANDLE destination, HANDLE _stdhandle,
...
+
+   Remarks:
+
+     - The lock is acquired and released within a single C call, because SRW
+       locks must be released by the OS thread that acquired it (a Haskell
+       thread may migrate between OS threads across FFI calls).
+     - We use 'safe' FFI calls for process spawning: an unsafe call blocked on
+       the lock would keep its capability, stalling every Haskell thread at
+       the next GC sync.
+     - SRW locks are neither fair nor FIFO, so there is no guarantee that a
```
Mistuke, line 373: Yeah, I think in practice it should be OK.  If we hit this issue then we can add a small FIFO queue on top of https://learn.microsoft.com/en-us/windows/win32/api/ntdef/ns-ntdef-list_entry

Mistuke (commented):
Thanks! This looks good. I can't approve though you need @snoyberg

